PR #16, opened 2025-03-28
Fix active run popup client

guillaq:
Closes https://linear.app/workflowai/issue/WOR-3660/active-runs

needs https://github.com/WorkflowAI/WorkflowAI/pull/15

Screenshot 2025-03-28 at 9 11 41 AM


anyacherniss:
This passes QA 👍 

Didn't check the exact count for all active tasks because some had hundreds in the past few days, but all the ones I counted exactly matched, and didn't spot any with 0 active runs 

CleanShot 2025-03-28 at 11 51 52@2x

PR #17, opened 2025-03-28
Correct model used by schema builder

yannbu:
Closes [WOR-4112: Validation errors triggered by the schema builder](https://linear.app/workflowai/issue/WOR-4112/validation-errors-triggered-by-the-schema-builder)

@guillaq there was a "typo" in my model picking for [WOR-4068: features/schemas triggers rate limits errors (Claude 3.7S)](https://linear.app/workflowai/issue/WOR-4068/featuresschemas-triggers-rate-limits-errors-claude-37s)

### DIFF
--- a/api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
+++ b/api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
@@ -42,7 +42,7 @@ class SchemaBuilderOutput(BaseModel):
 
 
 # TODO: switch back to Claude 3.7 when we'll have more token quotas
-@workflowai.agent(id="agent-schema-generation", model=workflowai.Model.CLAUDE_3_5_HAIKU_20241022)
+@workflowai.agent(id="agent-schema-generation", model=workflowai.Model.CLAUDE_3_5_SONNET_20241022)
 async def run_agent_schema_generation(
     input: SchemaBuilderInput,
 ) -> SchemaBuilderOutput:


guillaq (approved):

api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
```diff
@@ -42,7 +42,7 @@ class SchemaBuilderOutput(BaseModel):
 
 
 # TODO: switch back to Claude 3.7 when we'll have more token quotas
```
pierrevalade, line 44: is there a Linear issue to keep-track of this todo? @yannbu (should be linked here, with a due-date)

api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
```diff
@@ -42,7 +42,7 @@ class SchemaBuilderOutput(BaseModel):
...
-@workflowai.agent(id="agent-schema-generation", model=workflowai.Model.CLAUDE_3_5_HAIKU_20241022)
+@workflowai.agent(id="agent-schema-generation", model=workflowai.Model.CLAUDE_3_5_SONNET_20241022)
```
pierrevalade, line 45: why not use `LATEST` @yannbu?

api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
yannbu: https://linear.app/workflowai/issue/WOR-4123/revert-featuresschemas-to-claude-37s-when-well-have-higher-rate-limes

api/api/tasks/chat_task_schema_generation/schema_generation_agent.py
yannbu: LATEST models often suffer outtages in the first weeks after release, especially Anthropoc ones, this happened in the past with Claude 3.5 v2.... I'd rather handle the model updates myself.

guillaq: @yannbu `LATEST` points to OUR latest not the actual latest. 

> I'd rather handle the model updates myself.

That's exactly why we update the latest models manually........

yannbu: So if Claude 3.7 v2 is released, and stability is not good enough for X weeks, we'll still map "latest" to Claude 3.7 v1 for weeks ? How do you explain that to the users ?<br><br>I just find useful to decouple what models we internally think are ready for our production VS what our user see as the "latest".

Indeed we don't want to push a "latest" to users that is completely unstable but there might be a difference between what we consider internally is good enough for "latest" and what we want to communicate on the outside, does that make sense?

pierrevalade: i agree with @yannbu's point.

yannbu: by "unstable" I meant the model is not answering at all and the provider returns a 500. This was the case with Claude 3.5 Sony V2 if I recall correctly.Other instabilities can be generation instability

pierrevalade: OK

PR #19, opened 2025-03-28
Add minio storage for open source

guillaq:
Closes https://linear.app/workflowai/issue/WOR-3966/s3-compatible-storage


api/api/services/file_storage.py
```diff
@@ -0,0 +1,40 @@
...
+import os
+from typing import override
+
+from core.storage.file_storage import FileData, FileStorage
+
+
+def _default_file_storage() -> FileStorage:
+    # We return an empty blob storage if the env variables are not set
+
+    connection_string = os.getenv("WORKFLOWAI_STORAGE_CONNECTION_STRING", "")
```
yannbu, line 11: nit: use a `WORKFLOWAI_STORAGE_KIND` to make choices more explicit ?

api/core/storage/file_storage.py
```diff
@@ -0,0 +1,15 @@
...
+    content_type: str | None = None
+    filename: str | None = None
+
+
+class CouldNotStoreFileError(Exception):
+    pass
+
+
+class FileStorage(Protocol):
+    async def store_file(self, file: FileData, folder_path: str) -> str: ...
```
yannbu, line 15: just curious why do you use a Protocol this time and not an abstract class ?

api/core/storage/s3/s3_file_storage.py
```diff
@@ -0,0 +1,63 @@
...
+    async def store_file(self, file: FileData, folder_path: str) -> str:
+        # Generate a unique filename using content hash
+        content_hash = hashlib.sha256(file.contents).hexdigest()
+        extension = mimetypes.guess_extension(file.content_type) if file.content_type else None
+        key = f"{folder_path}/{content_hash}{extension or ''}"
+
+        try:
+            await asyncio.get_running_loop().run_in_executor(
+                None,
+                self._put_object,
```
yannbu, line 50: nit: I would explain why you had to use `asyncio.get_running_loop().run_in_executor` (boto3 is not async right?)

yannbu (approved):
Thank you! A few non-blocking comments.

api/core/storage/file_storage.py
guillaq: I think it's safe to use `Protocol` instead of ABC most times. The only upside of ABC is that you can add default implementation for functions (and inference at run time) but here there is no point and I only care about Protocol conformance.

`Protocol` has the benefit of not needing explicit subclassing which can be very useful.

api/api/services/file_storage.py
guillaq: That's what the scheme the connection string is for :) 
We already have a LOT of environment variables...

api/core/storage/s3/s3_file_storage.py
guillaq: Will merge now and tackle in a next PR !

PR #20, opened 2025-03-28
[WOR-3837] Logged-in new agent creation screen

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-3837/logged-in-new-agent-creation-screen-2-24

**Demo:**
Screen Recording 2025-03-28 at 4 36 59 PM


guillaq (approved):

anyacherniss:
**QA is done**

- [ ] Is it possible to keep the full screen chat when the user enters a message that doesn't immediately generate a schema?
CleanShot 2025-03-28 at 14 23 27@2x

- [ ] When I press "enter" to submit the feature description, the text field in the chat has an extra line added


https://github.com/user-attachments/assets/e0a727e4-d0ce-4365-a677-755c13f8f838

- [ ] Not seeing the expected response from the chat agent when creating a new schema. Not sure if the message is hardcoded in this PR or something? Expecting the same template message that's shown in the app right now

https://github.com/user-attachments/assets/45d33b4c-d87b-4f74-ad6c-0d472047469e

Current message shown in the app

CleanShot 2025-03-28 at 14 29 58@2x


- [ ] When selecting a prompt from the customer URL section, no message is shown as sent from the user. Not sure if there is something missing on the backend? 

(see second half of screen recording above)

- [ ] Would be great if we could support new lines/expanding the text field a bit (like the current UX). Not sure if it would look weird with the new UI though 😬 , but worth trying if it doesn't take too long? 


https://github.com/user-attachments/assets/30d6f192-d8a5-4ea2-b0ba-2fc4e00f0650

anyacherniss:
> @ anyacherniss Please confirm that’s the preferred solution, if not we need to add them to the backend response.

This works for me. Thank you @jacekzimonski!

---

**QA is done**

- [ ] When the message is multiple lines, can the button stay at the bottom? (see current button placement). 

CleanShot 2025-03-31 at 09 37 39@2x
CleanShot 2025-03-31 at 09 37 33@2x

- [ ] Subtle, but when I enter a multiline prompt, the textbox and sidebar text become a bit blurred. Really hard to see in the video, but it's more obvious in reality. 


https://github.com/user-attachments/assets/b358f5b4-9588-4099-99af-264fb1900017

anyacherniss:
This passes QA 👍

PR #23, opened 2025-03-31
[WOR-4122] Schemas Preview is called without schema $defs

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4122/schemaspreview-is-called-without-schema-dollardefs

@yannbu 
As discussed during our call we are sending it when it's defs that we are getting from chat.
It's probably a good idea to assign to you to QA, you know already what is happening and stuff.
@anyacherniss OK with you for @yannbu to QA?

Demo:
Screenshot 2025-03-31 at 12 14 17 PM
Screenshot 2025-03-31 at 12 12 45 PM

PR #24, opened 2025-03-31
handle missing defs in agent preview generation.

yannbu:
Closes https://linear.app/workflowai/issue/WOR-4120/error-with-preview-endpoint-an-files

Goes with https://github.com/WorkflowAI/WorkflowAI/pull/23
I initially thought that the frontend change would be enough but backend changes are required since in the case where the user add a document / image field itself, the frontend does not have the `$defs`.

### DIFF
--- a/api/api/services/internal_tasks/internal_tasks_service.py
+++ b/api/api/services/internal_tasks/internal_tasks_service.py
@@ -119,7 +119,7 @@ from core.tasks.task_input_example.task_input_migration_task import (
     stream_task_input_migration_task,
 )
 from core.tools import ToolKind
-from core.utils.schema_sanitation import add_missing_defs, streamline_schema
+from core.utils.schema_sanitation import streamline_schema
 from core.utils.schemas import EXPLAINATION_KEY, schema_needs_explanation, strip_json_schema_metadata_keys
 from core.utils.url_utils import extract_and_fetch_urls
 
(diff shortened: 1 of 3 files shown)


api/api/services/internal_tasks/internal_tasks_service.py
```diff
@@ -1103,38 +1103,50 @@ def stream_generate_task_preview(
...
+        self,
+        task_input: GenerateTaskPreviewTaskInput,
+        schema: dict[str, Any],
+        value: dict[str, Any],
+        error_attr: str,
+    ) -> None:
         try:
-            agent_output = SerializableTaskIO(
-                json_schema=task_input.task_output_schema,
+            # In some cases, $defs are missing from the schema sent by the frontend.
```
yannbu, line 1140: @guillaq to answer https://linear.app/workflowai/issue/WOR-4120/error-with-preview-endpoint-an-files#comment-b7852191.

api/core/utils/schema_sanitation.py
```diff
@@ -113,7 +113,7 @@ def model_schema(self, *args: Any, **kwargs: Any):
         return generated
 
 
-def _add_missing_defs(schema: dict[str, Any]) -> dict[str, Any]:
+def add_missing_defs(schema: dict[str, Any]) -> dict[str, Any]:
```
guillaq, line 116: just use `streamline_schema` ?

guillaq: This code is deprecated and should be removed soon

yannbu: OK I've added a deprecated decorator and migrated to streamline_schema.

guillaq (approved):

PR #29, opened 2025-04-01
What's next (after playground)

yannbu:
Related to [WOR-4022: Playground chat documentation: "what's next?"](https://linear.app/workflowai/issue/WOR-4022/playground-chat-documentation-whats-next)

Quite many added context and instructions to the Meta agent at @guillaq . Let me know if you have any questions.


api/api/services/internal_tasks/meta_agent_service.py
```diff
@@ -419,13 +420,28 @@ async def _build_meta_agent_input(
...
             self.fetch_agent_runs(task_tuple, playground_state.agent_run_ids),
+            self.feedback_service.list_feedback(
+                task_tuple[1],
+                run_id=None,
+                limit=10,
+                offset=0,
+                map_fn=MetaAgentInput.AgentLifecycleInfo.FeedbackInfo.AgentFeedback.from_domain,
+            ),
+            self.has_active_agent_runs(task_tuple, agent_schema_id),
+            self.get_reviewed_input_count(task_tuple, agent_schema_id),
```
guillaq, line 442: `return_exception=True` ? 

Otherwise you might end up with all call failing when a single one fails

guillaq (approved):
looks ok, careful with the asyncio.gather as it can be quite dangerous

api/api/services/internal_tasks/meta_agent_service.py
yannbu: @guillaq I've implemented a safer context fetching in this [commit](https://github.com/WorkflowAI/WorkflowAI/pull/29/commits/0b9dbff36ed6c935d0391e6a2588454f09484718). Do you want to check again? 

thanks !

guillaq (approved):
Might be better to make the code a bit DRYer but good for now !

PR #34, opened 2025-04-02
tune feature suggestion agent, add caching, llm as judge

yannbu:
Related to [WOR-3666: Suggested Agents Relevancy (based on company URL)](https://linear.app/workflowai/issue/WOR-3666/suggested-agents-relevancy-based-on-company-url)

@guillaq ideally I would like to put this feature in the hand of Anya and  Pierr today. If there are major comments, especially for the stream caching, maybe you can tackle that in subsequent MR tomorrow. I think we are fine since the whole thing is wrapped in a try-except.  But I've teste the caching heavily and got not errors or whatever.

### DIFF
--- a/api/core/utils/redis_cache.py
+++ b/api/core/utils/redis_cache.py
@@ -7,6 +7,8 @@ import pickle
 from collections.abc import AsyncIterator
 from typing import Any, Callable, Optional, TypeVar
 
+import redis.asyncio as aioredis
+
 F = TypeVar("F", bound=Callable[..., Any])
 AG = TypeVar("AG", bound=Callable[..., AsyncIterator[Any]])
 
@@ -14,14 +16,15 @@ _logger = logging.getLogger(__name__)
 
 
 def get_redis_client() -> Any:
-    import redis.asyncio as aioredis
-
     try:
         async_cache: aioredis.Redis | None = aioredis.from_url(os.environ["JOBS_BROKER_URL"])  # pyright: ignore
     except (KeyError, ValueError):
         async_cache = None
 
     return async_cache
+
+
+shared_redis_client: aioredis.Redis | None = get_redis_client()
 
 
 def _generate_cache_key(
@@ -50,8 +53,7 @@ def _generate_cache_key(
 
 
 def redis_cached(expiration_seconds: int = 60 * 60 * 24) -> Callable[[F], F]:  # default ttl is 1 day
-    async_cache = get_redis_client()
-    if not async_cache:
+    if not shared_redis_client:
         _logger.warning("Redis cache is not available, skipping redis_cached")
 
         def decorator(func: F) -> F:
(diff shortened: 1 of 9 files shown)


api/core/utils/redis_cache.py
```diff
@@ -57,3 +80,88 @@ async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    try:
+        cached_bytes = await async_cache.get(cache_key)  # pyright: ignore
+        if cached_bytes:
+            return pickle.loads(cached_bytes)  # pyright: ignore
+    except Exception as e:
+        _logger.exception("Failed to get cache for", exc_info=e, extra={"cache_key": cache_key})
+    return None
+
+
+async def _try_cache_result(async_cache: Any, cache_key: str, result: Any, expiration_seconds: int) -> None:
```
guillaq, line 96: Why `async_cache: Any` ?

api/core/utils/redis_cache.py
```diff
@@ -57,3 +80,88 @@ async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                        extra={"cache_key": cache_key},
+                    )
+
+            except Exception as e:
+                _logger.exception("Error in cached generator for", exc_info=e)
+                # Fallback to original function on error
+                async for item in func(*args, **kwargs):
+                    yield item
+            finally:
+                await _safely_close_redis(async_cache)
```
guillaq, line 162: does that work ?

You open the client once but close the connection after each call. Does redis automatically re-open connections when needed ? Seems like you should just be using a single, shared client ?

guillaq (approved):
You can go ahead and merge if that's urgent. But I am not 100% sure you will get cache hits since you open the redis client once and close it on every request ?

yannbu:
> You can go ahead and merge if that's urgent. But I am not 100% sure you will get cache hits since you open the redis client once and close it on every request ?

Cache works, 100%. That does not change the fact that the cache value is persisted. But OK I'll implemented a shared client. thanks!

api/core/utils/redis_cache.py
yannbu: shared client implemented here https://github.com/WorkflowAI/WorkflowAI/pull/34/commits/235c132ecf0f9d078e1b56441212ffa5f1158530

api/core/utils/redis_cache.py
yannbu: fixed https://github.com/WorkflowAI/WorkflowAI/pull/34/commits/121d613f6709abee9999cb4fd84564da838eebbe

api/core/utils/redis_cache.py
yannbu: Implement a shared cache in https://github.com/WorkflowAI/WorkflowAI/pull/34/commits/235c132ecf0f9d078e1b56441212ffa5f1158530

guillaq:
> Client open/close does not change the fact that the cache value is persisted, right? But OK I'll implemented a shared client. thanks!

No, but it's weird that after the first call you are essentially using a closed client to fetch and write to the cache. Maybe the redis library automatically re-opens the client if needed ?

guillaq (dismissed):

guillaq (approved):

PR #40, opened 2025-04-03
[WOR-4090] Generate input button shown/content "generated" for audi…

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4090/generate-input-button-showncontent-generated-for-audio-document

**Screenshot:**
Screenshot 2025-04-03 at 12 00 37 PM

### DIFF
--- a/client/src/lib/schemaFileUtils.ts
+++ b/client/src/lib/schemaFileUtils.ts
@@ -66,13 +66,7 @@ export function requiresFileSupport(
     return false;
   }
 
-  if (
-    schema.followedRefName === 'Image' ||
-    schema.followedRefName === 'Audio' ||
-    schema.followedRefName === 'Video' ||
-    schema.followedRefName === 'Document' ||
-    schema.followedRefName === 'File'
-  ) {
+  if (schema.followedRefName && FILE_REF_NAMES.includes(schema.followedRefName)) {
     return true;
   }
 


client/src/lib/schemaFileUtils.ts
```diff
@@ -58,7 +58,7 @@ export function replaceFileData(
   return innerReplaceFileDataWithURL(schema, schema.$defs, obj, replacer) as Record<string, unknown>;
 }
 
-export type FileFormat = 'audio' | 'image' | 'text' | 'document';
+export type FileFormat = 'audio' | 'image' | 'text' | 'document' | 'universal';
```
guillaq, line 61: Euh, why ? 
That's not one we support in the backend and will lead to errors no ?

Maybe `undefined` ? or you should strip it before sending it with the schema ?

client/src/lib/schemaFileUtils.ts
```diff
@@ -58,66 +58,46 @@ export function replaceFileData(
...
-      case 'pdf':
-      case 'text':
-      case 'document':
-        formats.add('document');
-        break;
-      default:
-        // TODO: how should we handle a file without a format?
-        break;
-    }
+  if (
```
guillaq, line 69: You should modify `FILE_REF_NAMES` to add `Document` and do `schema.followedRefName in FILE_REF_NAMES`

Also I don't think the Document ref name actually exists ?

guillaq (approved):

anyacherniss:
This passes QA 👍

PR #44, opened 2025-04-04
feat: use new payment endpoints

guillaq:


anyacherniss:
Unable to test because I can't add a card on this branch. Appears to be a regression because I'm able to add the same card on dev!


https://github.com/user-attachments/assets/f7ba60f8-8047-4213-adc5-2531e597dd82

guillaq:
Ah that's because the stripe config of the preview is messed up. I'll fix now, sorry about that

anyacherniss:
**QA is done**

- [ ] after adding a card that doesn't fail and reenabling auto recharge, I'm not charged. Not sure if that's in the scope of this PR, but wanted to verify because we definitely need to have it fixed! (Manually adding credits removed the failed state as expected though)


https://github.com/user-attachments/assets/d2c4b13a-ac00-4f2e-889a-0c0b6b995a4b



✅ adding test card works
✅ auto recharge failed state shows

anyacherniss:
@guillaq auto recharge be automatically disabled once it fails the first time? (Originally had it in a separate, frontend issue, but Pierre advised to have it handled by the backend)

guillaq:
> after adding a card that doesn't fail and reenabling auto recharge, I'm not charged. Not sure if that's in the scope of this PR, but wanted to verify because we definitely need to have it fixed! (Manually adding credits removed the failed state as expected though)

That's a backend issue. Yeah indeed we should fix ! I'll patch in a separate PR tonight

PR #47, opened 2025-04-07
Invalid unicode characters

guillaq:
ref https://linear.app/workflowai/issue/WOR-4065/issue-with-models-returning-invalid-unicode-characters


yannbu (approved):
thanks !

PR #48, opened 2025-04-07
[WOR-4154] [WOR-4155] Empty Reviews Page + Empty Benchmark Page

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4155/empty-benchmark-page-state
https://linear.app/workflowai/issue/WOR-4154/empty-reviews-page-state

**Screenshots:**


pierrevalade:
great work!

anyacherniss:
Passes QA for me too 👍 

(Pierre, as discussed on slack, I agree we should add additional documentation once I've created it, but I recommend we merge this as is since it's already an improvement and I'll create a new spec to add the links once they are created)

pierrevalade:
100%.

guillaq (approved):

PR #54, opened 2025-04-08
Remove spam warning

guillaq:


.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
+name: Release
+
+on:
```
yannbu, line 3: @guillaq those changes do not align with the MR title, were they introduced by accident ?

yannbu: other than that LGTM

PR #68, opened 2025-04-09
reworked stream_task_input_example_task instructions

yannbu:
Closes [WOR-4196: Duplicate email in generated input](https://linear.app/workflowai/issue/WOR-4196/duplicate-email-in-generated-input)


guillaq (commented):
@yannbu could we also handle WorkflowAIErrors ? otherwise the call never completes client side

PR #79, opened 2025-04-09
Cleanup json dict in runner

guillaq:
There was an issue with the test data. Unicode characters in JSON are double encoded meaning that it is the literal character "\" followed by the unicode hex value. So it looks like `\\u...` when serialized which I missed.

https://linear.app/workflowai/issue/WOR-4065/issue-with-models-returning-invalid-unicode-characters


yannbu (approved):
LGTM, thanks !

PR #81, opened 2025-04-10
use items instead of item_type in schema

yannbu:
Closes [WOR-4202: Investigate generation error of the agent builder](https://linear.app/workflowai/issue/WOR-4202/investigate-generation-error-of-the-agent-builder)

This is a simple change but sneaky to find. It seemed that the `item_type` naming was confusing the model into outputting "string" instead of {"type": "string"}. 

I now use "items" which align with the JSON schema standards and is less ambiguous.

I don't think we need to hotfix since errors are unfrequent. I'd rather have some time in staging to observe if we get more generation errors. But by generating all the schemas for the landing page, I didn't get errors, so that looks really good.

Can't wait for struct gen at Anthropic...


yannbu:
@guillaq can you check please ?

PR #89, opened 2025-04-10
Extract uptime for OpenAI and WorkflowAI

yannbu:
Closes [WOR-4168: API: WorkflowAI down time/up time in mins](https://linear.app/workflowai/issue/WOR-4168/api-workflowai-down-timeup-time-in-mins)
Closes [WOR-3209: API: OpenAI down time in mins](https://linear.app/workflowai/issue/WOR-3209/api-openai-down-time-in-mins)

Sadly, scrapping the Chat API uptime directly from https://status.openai.com/ is not easily with ScrapingBee, Firecrawl, (maybe https://browser-use.com/ would have helped), only the "all APIs" uptime is easily available without "clicking"

I might have gotten a little bit carried away with custom OpenAI scrapper, but at the same time if we don't display the exact uptime for the chat API, the feature doesn't make sense (that would be comparing apples and oranges: our /run api vs all OpenAI's APIs)

### DIFF
--- a/api/api/services/uptime_service.py
+++ b/api/api/services/uptime_service.py
@@ -132,7 +132,7 @@ class UptimeService:
             "Extract the API uptime",
         )
 
-        # BetterStack should display on time from 90 days from now. So we allow a small difference of 5 days
+        # BetterStack should display uptime from 90 days from now. So we allow a small difference of 5 days
         if since is not None:
             self._check_date_diff(since, date.today() - timedelta(days=90), 5, "workflowai")
         return UptimeInfo(uptime=uptime, since=since, source=URL)
@@ -152,6 +152,6 @@ class UptimeService:
             "Extract the 'APIs' uptime, DO NOT EXTRACT the uptime for 'ChatGPT'",
         )
         if since is not None:
-            # On this degraded scraping mode we can only know the month of start of the obtained data so we allow 32 days of tolerance.
+            # On this degraded scraping mode we can only know the month of start of uptime so we allow 32 days of tolerance.
             self._check_date_diff(since, date.today() - timedelta(days=90), 32, "openai")
         return UptimeInfo(uptime=uptime, since=since, source=URL)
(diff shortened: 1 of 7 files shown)


pierrevalade:
i think the other argument (other that simplifing) for using the "all APIs" data would be that now the uptime displayed on `status.openai.com` isn't the same we will display on our end (without tapping) — i would have picked the easier solution, feels like over-complexity.

yannbu:
Yes, but comparing `/run` uptime and all APIs (including fine-tune, embedding, sora, etc) just felt too much like comparing apples and oranges to me.<br><br>@pierrevalade can we merge in the current state still ?

pierrevalade:
yes!

api/core/agents/uptime_extraction_agent.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@workflowai.agent(
+    id="uptime-extraction-agent",
+    version=workflowai.VersionProperties(
+        model=workflowai.Model.GEMINI_2_0_FLASH_001,
+    ),
+)
+async def uptime_extraction_agent(input: UptimeExtractorAgentInput) -> workflowai.Run[UptimeExtractorAgentOutput]:
+    """Your specialist at extracting uptime values from status pages. You must extract the uptime from the 'status_page_content', and also take the eventual 'extraction_instructions' into account."""
```
guillaq, line 39:
```suggestion
    """You are a specialist at extracting uptime values from status pages. You must extract the uptime from the 'status_page_content', and also take the eventual 'extraction_instructions' into account."""
```

guillaq (approved):

api/api/services/uptime_service.py
```diff
@@ -0,0 +1,157 @@
...
+                "Uptime date difference is greater than tolerance",
+                extra={
+                    "from_date": from_date,
+                    "to_date": to_date,
+                    "tolerance": tolerance,
+                    "service_name": service_name,
+                },
+            )
+
+    async def _get_openai_chat_api_component_id(self) -> str | None:
```
guillaq, line 37: I am guessing the ID is not static ?

guillaq (approved):

api/api/services/uptime_service.py
yannbu: Not sure... so better be dynamic.

PR #91, opened 2025-04-11
double check when playground agent answers with ask_user_confirmation…

yannbu:
Closes https://linear.app/workflowai/issue/WOR-4219/playground-agent-problem-with-ask-user-confirmation=false

### DIFF
--- a/api/api/services/internal_tasks/meta_agent_service_test.py
+++ b/api/api/services/internal_tasks/meta_agent_service_test.py
@@ -804,3 +804,122 @@ class TestMetaAgentService:
 
             # For reviewed_input_count
             assert result.reviewed_input_count == expected_context.reviewed_input_count
+
+    @pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


api/api/services/internal_tasks/meta_agent_service_test.py
```diff
@@ -804,3 +804,122 @@ async def test_fetch_meta_agent_context(
 
             # For reviewed_input_count
             assert result.reviewed_input_count == expected_context.reviewed_input_count
+
+    @pytest.mark.asyncio
```
guillaq, line 808: not needed

guillaq (approved):

api/api/services/internal_tasks/meta_agent_service_test.py
yannbu: right deleted https://github.com/WorkflowAI/WorkflowAI/pull/91/commits/9ceef69a03d9b7a6ef529b80eb25af8cd6324485

PR #103, opened 2025-04-11
update playground agent docs

yannbu:
Closes [WOR-4228: Refresh playground agent docs (when Anya is done updating)](https://linear.app/workflowai/issue/WOR-4228/refresh-playground-agent-docs-when-anya-is-done-updating)


guillaq (approved):
Any chance you could move the docs dir at the root of the repo ? It would be clearer ? (don't forget to copy it in the right spot in the docker image)

PR #120, opened 2025-04-14
[WOR-4250] Don't scroll to top on: Workflowai.com/apple.com when changing search

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4250/dont-scroll-to-top-on-workflowaicomapplecom-when-changing-search

@pierrevalade @anyacherniss 
When we changed the redirect after entering the company URL from query parameter to new url (workflowai.com/apple.com), we started loading a new page. So the scroll automatically scrolls to top (becasue now it's a new url). 

To overcome this behaviour, we still use the company url from our url, but when the parameter changes we don't redirect to new url. 
- But made sure that the copy button still copies the correct url!
- It also does not scroll to top anymore

**Demo:**
Screen Recording 2025-04-14 at 5 50 22 PM

### DIFF
--- a/client/src/app/landing/URLLandingPage.tsx
+++ b/client/src/app/landing/URLLandingPage.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useCallback } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { LandingPageContainer } from './container/LandingPageContainer';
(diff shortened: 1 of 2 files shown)


guillaq (approved):

anyacherniss:
Updated scrolling behavior works, so I say this passes QA 👍 (the correct tasks are not displayed, but that's out of scope)

PR #121, opened 2025-04-14
Handle payment errors in playground

guillaq:
Screenshot 2025-04-14 at 6 13 07 PM
Screenshot 2025-04-14 at 6 17 44 PM

https://linear.app/workflowai/issue/WOR-3691/block-runs-when-balance-dollar0-for-an-org-starting-218

https://linear.app/workflowai/issue/WOR-4193/turn-off-auto-recharge-when-auto-recharge-fails


guillaq:
@anyacherniss merging so you only have to QA the release branch

PR #133, opened 2025-04-15
[WOR-4260] WorkflowAI.com/apple.com <- show only company specific entries

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4260/workflowaicomapplecom-show-only-company-specific-entries

**Demo:**
Screen Recording 2025-04-15 at 2 28 58 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -75,5 +75,6 @@
   },
   "[yaml]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "jest.rootPath": "client"
 }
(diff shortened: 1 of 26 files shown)


guillaq (approved):

anyacherniss:
CleanShot 2025-04-15 at 09 12 40@2x
When I enter a different URL, on the webpage, the page URL is not updated. Unclear if that's a blocker or not. @pierrevalade does the existing behavior (URL doesn't update) allow you to do what you need to do, or does this problem need to be fixed?

anyacherniss:
Confirmed on slack the previous issue is not a blocker

PR #140, opened 2025-04-16
[WOR-4275] New Button: Logged-in homepage button to open the API key modal

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4275/new-button-logged-in-homepage-button-to-open-the-api-key-modal

**Screenshots:**
Screenshot 2025-04-16 at 10 06 06 AM
Screenshot 2025-04-16 at 10 06 09 AM


guillaq (approved):
Code looks ok. I would expect this button to be in the user menu though ? (@anyacherniss )

anyacherniss:
This passes QA 👍 

I think the positioning is fine for now! Based on the original request on slack, it sounds like we want the button directly on the page
> please add button tomorrow on our logged-in homepage to open the API key modal?

It's probably not going to be permanent positioning anyway, but it's great for a quick fix. For context, _eventually_ we're adding a new menu on the logged in home page for org-wide features, but that would only make sense once additional features are built (so it's not a priority at this time).

PR #149, opened 2025-04-17
[WOR-4280] HTML streaming issues

jacekzimonski:
**Closes:**
https://linear.app/workflowai/issue/WOR-4280/html-streaming-issues

**Demo:**
Screen Recording 2025-04-17 at 1 48 30 PM


guillaq (approved):
Pretty smart I think !
I was afraid you would miss the last value but since `isRendering` will change you should be good. 

The one think that's missing is a `useIsMounted` to make sure you don't call the state changes if the component is no longer mounted

pierrevalade:
look who is becoming the react master

PR #153, opened 2025-04-17
Deprecate old llama models

guillaq:
ref https://linear.app/workflowai/issue/WOR-4282/fireworks-models-getting-undeployed

### DIFF
--- a/api/core/domain/models/_mistral.py
+++ b/api/core/domain/models/_mistral.py
@@ -8,24 +8,8 @@ from core.domain.models.providers import Provider
 
 def mistral_models() -> dict[Model, ModelData | LatestModel | DeprecatedModel]:
     return {
-        Model.MIXTRAL_8X7B_32768: ModelData(
-            display_name="Mixtral (8x7B)",
-            supports_json_mode=True,
-            supports_input_image=False,
-            supports_multiple_images_in_input=False,
-            supports_input_pdf=False,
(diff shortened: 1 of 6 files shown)


api/core/domain/models/_mistral.py
```diff
@@ -8,24 +8,8 @@
...
-                max_tokens=32768,
-                source="https://api.fireworks.ai/v1/accounts/fireworks/models/mixtral-8x7b-instruct",
-            ),
-            provider_for_pricing=Provider.FIREWORKS,
-            icon_url="https://workflowai.blob.core.windows.net/workflowai-public/mistral.svg",
-            release_date=date(2023, 12, 11),
-            quality_index=468,  # MMLU=70.60, GPQA=39.00
-            provider_name=DisplayedProvider.MISTRAL_AI.value,
-            supports_tool_calling=False,
+        Model.MIXTRAL_8X7B_32768: DeprecatedModel(
```
yannbu, line 11: nit: we could fallback to MISTAL_LARGE to have a higher quality index instead of lower ? (MISTAL smalli is lower than Mixtral)

yannbu (approved):

api/core/domain/models/_mistral.py
guillaq: Actually no :/ Mistral large is 10x more expensive...

PR #158, opened 2025-04-17
Order Should Matter When Editing New Schema

jacekzimonski:
@guillaq @anyacherniss @pierrevalade 
So implemented checking of the order for Property Keys in Schema.
On frontend it's working.
But after saving the modified schema, backend informs us that the schema already exists and redirects to the one with old order.
So the order check to make it work should be also implemented on the backend to make it work.
This PR has the required Frontend part.
@guillaq was informed

Screen Recording 2025-04-17 at 10 12 47 PM


guillaq:
-> https://linear.app/workflowai/issue/WOR-4368/order-should-matter-when-editing-schema

PR #172, opened 2025-04-21
feat: add groq provider for llama4 models

guillaq:
ref https://linear.app/workflowai/issue/WOR-4338/llama4-through-groq


pierrevalade:
oh oh nice! 💌

PR #176, opened 2025-04-22
CSM agent feature

yannbu:
related to [WOR-4271: CSM Agent](https://linear.app/workflowai/issue/WOR-4271/csm-agent)

The first slice of CSM Agent. It's still a work in progress, but I would like to deploy early, at least in staging, to start setting up the whole flow with the webhook on staging, etc.

### DIFF
--- a/api/core/utils/redis_cache.py
+++ b/api/core/utils/redis_cache.py
@@ -76,7 +76,7 @@ def _generate_cache_key(
 
 
 def redis_cached(expiration_seconds: int = 60 * 60 * 24) -> Callable[[F], F]:  # noqa: C901
-    def decorator(func: F) -> F:
+    def decorator(func: F) -> F:  # noqa: C901
         @functools.wraps(func)
         async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
             if not shared_redis_client:


api/api/main.py
```diff
@@ -131,6 +133,7 @@ async def lifespan(app: FastAPI):
 
 app.include_router(probes.router)
 app.include_router(run.router)
+app.include_router(webhooks.router)
```
guillaq, line 136: Careful. Use the same pattern as the clerk and stripe webhooks in `main_router.py` otherwise the endpoint will be included in the run endpoint

api/api/tags.py
```diff
@@ -22,3 +22,4 @@ class RouteTags(str, Enum):
     FEATURES = "Features"
     PROMPT_ENGINEER_AGENT = "Prompt Engineer Agent"
     FEEDBACK = "Feedback"
+    WEBHOOKS = "Webhooks"
```
guillaq, line 25: Can you add to the other webhooks then ?

api/api/routers/webhooks.py
```diff
@@ -0,0 +1,29 @@
+import logging
```
guillaq, line 1: rename file to `slack_webhooks`

api/core/storage/slack/slack_api_client.py
```diff
@@ -140,3 +162,86 @@ async def get_channel_info(self, channel_id: str):
...
+                    # This message is already being processed or was processed recently
+                    _logger.info("Skipping duplicate webhook event", extra={"client_msg_id": client_msg_id})
+                    return
+
+            await self._process_webhook_message(webhook_event)
+
+        except Exception as e:
+            _logger.exception("Failed to handle Slack webhook", extra={"payload": raw_payload, "error": e})
+
+    async def _process_webhook_message(self, webhook_event: SlackWebhookEvent) -> None:
```
guillaq, line 224: Careful, this is in the wrong layer. The slack client should only know about slack and not about our internal processes.

The easiest way to respect layer boundaries would be to have `handle_webhook` return a validated webhook event or None.
And move the below logic to a class method in the customer service that you can call in the router directly

api/core/utils/redis_lock.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Try to acquire the lock - simply set a key if it doesn't exist
+    acquired = await shared_redis_client.set(lock_key, "1", nx=True, ex=expire_seconds)
+
+    if not acquired:
+        _logger.info("Failed to acquire lock, resource already locked", extra={"lock_key": lock_key})
+        raise LockAcquisitionError(f"Could not acquire lock: {lock_key}")
+
+    try:
+        _logger.debug("Lock acquired", extra={"lock_key": lock_key})
```
guillaq, line 40: remove debug logs ? 
Also the name is a bit misguiding because you never actually unlock anything.
I would rename to `redis_dedup`

api/core/storage/slack/slack_api_client.py
yannbu: yes yes the business logic should go elsewhere. I just rushed to push the code.<br><br>But I can adjust.

api/api/main.py
yannbu: https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/cab0983ae3820fde93c46e1d15073b9924c93eb7

api/api/routers/webhooks.py
yannbu: https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/cab0983ae3820fde93c46e1d15073b9924c93eb7

api/api/tags.py
yannbu: After thinking about it, it doesn't make sense if we don't display the webhooks in the docs.

api/core/storage/slack/slack_types.py
```diff
@@ -87,19 +89,17 @@ class SlackEventData(NamedTuple):
...
+class SlackAuthorization(BaseModel):
     team_id: str
     user_id: str
     is_bot: bool
     is_enterprise_install: bool
+    enterprise_id: str | None = None
 
 
-class SlackWebhookEvent(NamedTuple):
+class SlackWebhookEvent(BaseModel):
```
yannbu, line 100: @guillaq the nested parts of the SlackWebhookEvent are parsed as dict instead of NamedTuple, which breaks the code so I'll stick to BaseModel for now.

yannbu: done in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/e69a20d7b23e909eaf380e9a569deadf2fcce8b5

api/core/storage/slack/slack_api_client.py
yannbu: done in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/e69a20d7b23e909eaf380e9a569deadf2fcce8b5

api/api/routers/helpscout_webhooks.py
```diff
@@ -0,0 +1,78 @@
...
+        payload = await request.json()
+        _logger.info("Received Helpscout webhook", extra={"payload": payload})
+        event = RawWebhookEvent.model_validate(payload)
+        if event.type != "email":
+            _logger.info("Received non-email webhook event, skipping", extra={"event": event})
+            return {"status": "success"}
+
+        email = event.to_email()
+
+        await CustomerService.handle_helpscout_email_sent(email, storage)
```
yannbu, line 73: TODO: put in `add_background_task`, but @guillaq I notice we don't get the logs for code that run in `add_background_task` did you notice that too ?

guillaq: Ah no I didn't. Are you sure ? I can try again later today

api/api/services/features.py
```diff
@@ -402,13 +412,14 @@ async def stream_features_by_domain(
         # Wait for AI features collection to complete
         company_website_contents = await company_website_contents_task
 
-        agent_suggestion_input = await self._build_agent_suggestion_input(
+        agent_suggestion_input = await FeatureService._build_agent_suggestion_input(
```
guillaq, line 415:
```suggestion
        agent_suggestion_input = await cls._build_agent_suggestion_input(
```

api/core/services/users/clerk_user_service.py
```diff
@@ -72,6 +73,26 @@ async def get_org_admins(self, org_id: str, max_users: int = 5) -> list[UserDeta
                 return []
             return await self._get_users_by_id(client, user_ids)
 
+    async def get_user_id_by_email(self, email: str) -> str:
```
guillaq, line 76: Can we cache the result ? Clerk rates limit are very small

api/core/services/users/clerk_user_service.py
```diff
@@ -72,6 +73,26 @@ async def get_org_admins(self, org_id: str, max_users: int = 5) -> list[UserDeta
...
+        data: list[dict[str, Any]] = response.json()
+
+        if len(data) == 0:
+            raise ObjectNotFoundError(f"No user found for email {email}")
+        if len(data) > 1:
+            raise InternalError(f"Several users found for email {email}")
+
+        return data[0]["id"]
+
+    async def get_user_organization_ids(self, user_id: str) -> list[str]:
```
guillaq, line 89: same here

api/core/storage/mongo/partials/mongo_organizations.py
```diff
@@ -404,6 +404,23 @@ async def feedback_slack_hook_for_tenant(self, tenant_uid: int) -> str | None:
             raise ObjectNotFoundException("Organization not found", code="organization_not_found")
         return doc.get("feedback_slack_hook")
 
+    @override
+    async def get_organization_by_slack_channel_id(self, slack_channel_id: str) -> TenantData | None:
+        cursor = self._collection.find({"slack_channel_id": slack_channel_id})
```
guillaq, line 409: We should index the field to avoid spamming the db. I'll do in a separate PR.

api/core/storage/mongo/partials/mongo_organizations.py
```diff
@@ -404,6 +404,23 @@ async def feedback_slack_hook_for_tenant(self, tenant_uid: int) -> str | None:
...
+        docs: list[OrganizationDocument] = []
+        async for doc in cursor:
+            docs.append(OrganizationDocument.model_validate(doc))  # noqa: PERF401
+
+        if not docs:
+            return None
+
+        if len(docs) > 1:
```
guillaq, line 417: Let's just make the index unique

api/core/storage/organization_storage.py
```diff
@@ -42,6 +42,8 @@ async def add_credits_to_tenant(self, tenant: str, credits: float) -> None: ...
 
     async def decrement_credits(self, tenant: str, credits: float) -> TenantData: ...
 
+    async def get_organization_by_slack_channel_id(self, slack_channel_id: str) -> TenantData | None: ...
```
guillaq, line 45: Do you need the full organization object ? I would move to the PublicOrganizationStorage and return a`PublicOrganizationData`.

api/core/storage/mongo/partials/mongo_organizations.py
guillaq: You should just use `_get_public_org` ?

guillaq (approved):
Some minor comments !

api/core/storage/mongo/partials/mongo_organizations.py
guillaq: -> https://github.com/WorkflowAI/WorkflowAI/pull/240

yannbu: thank you !

api/core/services/users/clerk_user_service.py
yannbu: Addressed in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/63e63a9ee9b79e05667a321b73e4b8ff7632dd33

This is a cache update because we want to ignore the self / cls arguments.

api/core/services/users/clerk_user_service.py
yannbu: Addressed in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/63e63a9ee9b79e05667a321b73e4b8ff7632dd33

This is a cache update because we want to ignore the self / cls arguments.

api/core/storage/mongo/partials/mongo_organizations.py
yannbu: I now use _get_public_org in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/63e63a9ee9b79e05667a321b73e4b8ff7632dd33

api/core/storage/organization_storage.py
yannbu: I use _get_public_ord in https://github.com/WorkflowAI/WorkflowAI/pull/176/commits/63e63a9ee9b79e05667a321b73e4b8ff7632dd33

api/core/services/users/clerk_user_service.py
yannbu: Do you want to have a quick look at the updated cache func ?

PR #190, opened 2025-04-24
Image generation

guillaq:


api/core/providers/openai_image/openai_image_domain.py
```diff
@@ -0,0 +1,75 @@
...
+    created: int = Field(description="The Unix timestamp of when the image was created.")
+
+    class Data(BaseModel):
+        url: Optional[str] = Field(default=None, description="The URL of the generated image.")
+        b64_json: Optional[str] = Field(default=None, description="The base64-encoded JSON of the generated image.")
+
+        def to_file(self, content_type: str) -> File:
+            return File(data=self.b64_json, content_type=content_type)
+
+    data: List[Data] = Field(description="The list of generated images.")
```
yannbu, line 61: nit: this is the old typing (List)

api/core/providers/openai_image/openai_image_provider.py
```diff
@@ -0,0 +1,168 @@
...
+            else "https://api.openai.com/v1/images/generations"
+        )
+
+        async with self._open_client(url) as client:
+            return await client.post(
+                url,
+                # Probably have to send the mask and image as files
+                headers={"Authorization": f"Bearer {self._config.api_key}"},
+                json=request.model_dump(mode="json", exclude_none=True, by_alias=True),
+                timeout=options.timeout,
```
yannbu, line 136: @guillaq which timeout value is used by default ? It should be higher for this provider.

yannbu (approved):

api/core/providers/openai_image/openai_image_domain.py
guillaq: ah yes. we should add it to cursor rules

api/core/providers/openai_image/openai_image_provider.py
guillaq: 3min

PR #207, opened 2025-04-25
[WOR-4211] Bug: same price

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4211/bug-same-price

Screenshot:
Screenshot 2025-04-25 at 4 17 24 PM

### DIFF
--- a/client/src/lib/formatters/numberFormatters.ts
+++ b/client/src/lib/formatters/numberFormatters.ts
@@ -45,6 +45,10 @@ export function formatFractionalCurrencyAsNumber(value: number | null | undefine
     return null;
   }
 
+  if (value === 0) {
+    return 0;
+  }
+
   if (Math.abs(value) >= 0.0001) {
     return Number(value.toFixed(4));
   }


guillaq (approved):

anyacherniss:
**QA is done**
- [ ] It's working when the lowest price is the same; still seeing mismatch when the matching prices are more expensive:

https://workflowai-ipcboqvxt-workflowai.vercel.app/workflowai/agents/animal-cuteness-and-danger-assessment/15?versionId=686608173a4e7c740f4240fecca71543&showDiffMode=false&show2ColumnLayout=false&taskRunId3=01967c8b-ff4f-71f5-104d-3c86c1e5fe8c&taskRunId1=01967c8b-ffbc-7381-ed0b-00b4053e82d8&taskRunId2=01967c8c-0009-73ee-9514-275b78bf2331

CleanShot 2025-04-28 at 09 17 58@2x
CleanShot 2025-04-28 at 09 17 37@2x

anyacherniss:
The price is the same but the multipliers are not. Same price but one is 4.4x more expensive and the other is 5.4x more expensive. Two matching prices should the same amount more expensive. 

CleanShot 2025-04-28 at 12 25 26@2x

guillaq:
I agree with Jacek on the fact that the x should be computed on exact and not rounded price

anyacherniss:
> I agree with Jacek on the fact that the x should be computed on exact and not rounded price

If that's the case, then we should adjust the behavior when the two "matching" (matching-when-rounded) prices are also the lowest. Right now there is an inconsistency, which is my issue. Based on the original bug issue that Pierre opened, my best guess in how to address the inconsistency was to calculate based on the prices as displayed. Thus, rounding.

If you both feel very strongly about calculating off of the exact price, then we should do it both when the price is higher and when the price is the lowest, and explain to Pierre why the behavior he flagged as a bug isn't a bug, and will continue.

anyacherniss:
**QA is done**

The price calculation looks good to me. It does looks like what I would expect based on the description of the original issue. 

- [ ] When running a free model, the app crashes

https://workflowai-bgaesq11h-workflowai.vercel.app/workflowai/agents/animal-cuteness-and-danger-assessment/15?showDiffMode=false&show2ColumnLayout=false&versionId=e6d698de387ab45b365cbd4c893e123c&taskRunId3=01967e4f-3791-73cf-096f-bfed1793663a&taskRunId1=01967e4e-1067-7326-9db0-5ab825748b04


https://github.com/user-attachments/assets/5791eff7-bbfa-4fa4-ac29-f860599a6360

anyacherniss:
This passes QA 👍

PR #209, opened 2025-04-25
[WOR-4401] Potential fix: Improve prompt error (but improve prompt is working)

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4401/improve-prompt-error-but-improve-prompt-is-working

Explanation:
@anyacherniss 
Not hundred percent sure if this fix will work, was not eable to replicate this issue.
But that's my best theory.

So:
- I think the issue was when the previous stream failed, but was not cleaned up, and did not show the prompt.
- When the new stream concluded, the previous stream reported error (the one from your previous instruction try), and reported it at the same time as the success
- To prevent that I'm doing two things:
- I'm canceling previous stream from chat as soon as new starts
- I added abortsignal also to instruction change from the outputs, and also canceling it


guillaq (approved):

anyacherniss:
This passes QA 👍

PR #213, opened 2025-04-28
Add Image Generation task in features section

yannbu:
Closes https://linear.app/workflowai/issue/WOR-4393/add-image-generation-task-in-features-section


yannbu:
@guillaq on sec, I need to change the image URL

PR #222, opened 2025-04-29
Gemini image gen

yannbu:
Screenshot 2025-04-29 at 11 54 16

### DIFF
--- a/api/core/runners/workflowai/workflowai_runner.py
+++ b/api/core/runners/workflowai/workflowai_runner.py
@@ -6,7 +6,7 @@ from collections.abc import Sequence
 from copy import deepcopy
 from typing import Any, Callable, Iterable, NamedTuple, Optional
 
-from pydantic import TypeAdapter, ValidationError
+from pydantic import TypeAdapter
 from typing_extensions import override
 
 from api.services.providers_service import shared_provider_factory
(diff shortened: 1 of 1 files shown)


api/core/runners/workflowai/workflowai_runner.py
```diff
@@ -418,27 +422,13 @@ async def _extract_image_options(
...
-            return base
-
-        # In case of validation, we can check which key was actually extracted
-        extracted_keys = extracted_options.model_dump(exclude_unset=True)
 
-        if delete_keys:
-            # We can remove the corresponding fields from the input since they are already used
-            for key in extracted_keys:
-                input.pop(key, None)
+        if base is None:
```
yannbu, line 426: @guillaq without that the image count is None, and this code is skipped:

`if messages[0].image_options and messages[0].image_options.image_count:
            generation_config.responseModalities = ["IMAGE", "TEXT"]`

yannbu: I'll let you see what you do with that, but I think there is no "image_options" in the input schema right ?

api/core/runners/workflowai/workflowai_runner.py
```diff
@@ -347,7 +347,11 @@ def _pick_template(
             # a bit hacky but we templates that have structured generation enabled do
             # not show the output schema
             is_structured_generation_enabled=is_structured_generation_enabled
-            or not self._prepared_output_schema.prepared_schema.get("properties"),
+            or (
```
yannbu, line 350: Without that change the wrong template is picked (struct gen), and since Gemini 2.0 does not use stuct gen, the model has zero info about what to generate. This is the reason why we had empty output.

guillaq:
As discussed, closing in favor of https://github.com/WorkflowAI/WorkflowAI/pull/223

PR #228, opened 2025-04-30
Basics for openai proxy

guillaq:


yannbu:
@guillaq will you storage a `is_from_proxy` flag like we do for `is_active` to spot runs that are triggered from the proxy ?

guillaq:
> @guillaq will you store a `is_from_proxy` flag like we do for `is_active` to spot runs that are triggered from the proxy ?

I was not going to ? I can add it to a metadata if you would like ? What would the distinction be for ?

yannbu:
Customer success, understanding how the user use our product (SDK / Api / proxy)

guillaq:
> Customer success, understanding how the user use our product (SDK / Api / proxy) 

I think metadata is good for that use case !

PR #254, opened 2025-05-08
[WOR-4465] Proxy Playground

jacekzimonski:
Changes to the proposed designs needed to be done.
Lot of shortcuts needed to be taken.
In this case we shouldn't stick to the screens on linear but more focus on the features.

All the changes were consulted with @guillaq to make most sense and to publish a version quickly.

There are still some missing parts like:
- We need to be able to create a Agent Message
- We need to modify the tool calls
- We need to include the ToolCalls in the reply functionality (otherwise they are missed in the next round)
- And bunch of tweaks that are smaller
But will tangle that tomorrow.

Becasue the changes will not change the main flow and construction, we still will benefit from Code Reviewing it and QAing.
Demo will be included and avaible in the zoom recording.


client/src/app/[tenant]/agents/[taskId]/[taskSchemaId]/playground/hooks/useIsProxy.tsx
```diff
@@ -0,0 +1,11 @@
+import { TaskSchemaResponseWithSchema } from '@/types/task';
+
+export function useIsProxy(schema: TaskSchemaResponseWithSchema) {
```
Comment (guillaq) [inline client/src/app/[tenant]/agents/[taskId]/[taskSchemaId]/playground/hooks/useIsProxy.tsx:3]: for input variables you'll have to check if there is a `messages` key in the version properties

anyacherniss:
Thanks @jacekzimonski!

The bugs that were expected to be fixed all look good. Regarding:

> I wonder if - to avoid the bounce when hovering over a message - we should the type selector over the message below it? Since the user would not need to interact with the user message below it at the same time anyway? (Not 100% convinced of this - open to feedback if you disagree)

Ignore this - we're having Victor work on some updated UI options, so this will probably not be relevant.

PR #263, opened 2025-05-13
[WOR-4464] Frontend Proxy Onboarding UX

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4464/frontend-proxy-onboarding-ux

Demo:
Screen Recording 2025-05-13 at 2 51 49 PM

### DIFF
--- a/client/src/app/[tenant]/agents/[taskId]/[taskSchemaId]/playground/playgroundOutput.tsx
+++ b/client/src/app/[tenant]/agents/[taskId]/[taskSchemaId]/playground/playgroundOutput.tsx
@@ -153,6 +153,7 @@ function ModelOutput(props: ModelOutputProps) {
             open={openModelCombobox}
             setOpen={setOpenModelCombobox}
             isProxy={isProxy}
+            taskId={taskId}
           />
           <CreateTaskRunButton
             taskRunner={taskRunner}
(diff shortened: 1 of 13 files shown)


anyacherniss:
@jacekzimonski took a couple notes during the demo on some polishes. I think the only item that would have a noticeable impact on users would be the lack of a copy button for the code. The rest is cosmetic/nonblocking and can be done later if not quick to update.


- [ ] Add a button to copy the generated code
CleanShot 2025-05-13 at 14 39 03@2x
- [ ] Use the instructor icon instead of Python icon. Icon can be exported from the figma for this UX - let me know if you have issues exporting it though, I can put it in a .zip for you
CleanShot 2025-05-13 at 14 35 05@2x
CleanShot 2025-05-13 at 14 37 44@2x
- [ ] Update code display UI to match colors of code in figma (UI doesn't need to match exactly, but would be great to at least use similar colors
CleanShot 2025-05-13 at 14 40 03@2x
CleanShot 2025-05-13 at 14 40 13@2x

yannbu (approved):

anyacherniss:
@jacekzimonski I still don't see the copy button, which I believe is frontend. Please confirm if otherwise. Your previous comment only acknowledges 2/3 items I listed in https://github.com/WorkflowAI/WorkflowAI/pull/263#issuecomment-2877592442.

Since we need it for the demo, if a preview branch cannot be made without merging this (I'm still getting confused with the new release process) it can be merged without the copy button, as long as a follow up is documented for that update!

PR #268, opened 2025-05-14
Proxy playground agent

yannbu:
@guillaq @jacekzimonski we'll need that for today's demo.


api/core/storage/clickhouse/clickhouse_client.py
```diff
@@ -522,19 +522,23 @@ async def run_count_by_agent_uid(
...
-    async def list_runs_since(
+    async def list_latest_runs(
         self,
-        since_date: datetime,
+        task_uid: int | None = None,
+        since_date: datetime | None = None,
         is_active: bool = True,
         limit: int = 100,
     ) -> AsyncIterator[AgentRun]:
-        # TODO: use SerializableTaskRunQuery instead
```
guillaq, line 531: is that no longer true ? 
I am not 100% sure this function is useful ? Might be good to avoid duplicating logic ?

api/core/domain/integration_domain_test.py
```diff
@@ -1,8 +1,9 @@
-from core.domain.integration_domain import (
+from core.domain.integration_domain.integration_domain import (
```
guillaq, line 1: this should be moved to the `integration_domain` dir ?

Also the directory should likely be named `integrations` ?

api/core/domain/integration_domain/openai_sdk_python_snippets.py
```diff
@@ -0,0 +1,73 @@
+OPENAI_SDK_PYTHON_LANDING_PAGE_SNIPPET = """import os
```
guillaq, line 1: I don't think this should be at the domain layer. Only the types should be here to avoid clutter ?

`OFFICIAL_INTEGRATIONS` should be at the service layer ?

api/core/domain/integration_domain/openai_sdk_ts_snippets.py
```diff
@@ -0,0 +1,92 @@
...
+  ],
+});
+
+console.log(response.choices[0].message.content);
+"""
+
+
+# Landing Page (Reliable Structured Output) -----------------------------------
+# Requires openai@^4.55.0 for `beta.chat.completions.parse`.
+OPENAI_SDK_TS_LANDING_PAGE_STRUCTURED_GENERATION_SNIPPET = """import OpenAI from 'openai';
```
guillaq, line 35: It would be better to extract to a separate file with a proper extension. This way we can use the editor auto-completion.

guillaq: something like `services/internals/snippets/...` ?

guillaq (approved):
Code is OK but the layers are a bit messy IMO. Ok to fix later given the rush...

api/core/domain/integration_domain/openai_sdk_ts_snippets.py
yannbu: I have thought about it and that would be super cool, but the problem is the import—for example, in Python we import instructor which is not in the code base—so it's breaking the linting.

api/core/domain/integration_domain/openai_sdk_python_snippets.py
yannbu: Okay, I've put those in integration mapping in the domain, because I think it's still the domain—some business config --, like we do for the models data, for example.

api/core/domain/integration_domain_test.py
yannbu: fixed

api/core/storage/clickhouse/clickhouse_client.py
yannbu: still true

api/core/domain/integration_domain/openai_sdk_ts_snippets.py
yannbu: resolving to merge

yannbu: but happy to discuss

guillaq: We could simply add instructor as a dev dependency ?

yannbu: Yes, but then you would end up in production with some Python code in the container that imports an unknown package. I mean, that might or might not crash the container, but feels a bit weird to me.

PR #272, opened 2025-05-14
[WOR-4670] Redirection bug in onboarding

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4670/redirection-bug-in-onboarding

Explanation:
On Preview a newer version with different redirect response from backend was deployed that is not compatible with the older one. This PR adjusts the redirect for the new structure.

Important:
It will work when the changes with redirect that are already avaible on Preview will be propagated to dev.
@yannbu will inform when.


yannbu (approved):

guillaq (approved):

guillaq:
Merging, don't think there is much to QA here

PR #278, opened 2025-05-15
Various enhancements on proxy playground agent

yannbu:
Closes [WOR-4672: Generated code in chat should match code in user's task](https://linear.app/workflowai/issue/WOR-4672/generated-code-in-chat-should-match-code-in-users-task)
Closes [WOR-4673: Generated code in chat should use the correct model](https://linear.app/workflowai/issue/WOR-4673/generated-code-in-chat-should-use-the-correct-model)
Closes [WOR-4674: instructions not properly formatted in messages @yann](https://linear.app/workflowai/issue/WOR-4674/instructions-not-properly-formatted-in-messages-yann)

### DIFF
--- a/api/core/domain/task_variant.py
+++ b/api/core/domain/task_variant.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel, Field, model_validator
 
 from core.domain.errors import JSONSchemaValidationError
 from core.domain.fields.chat_message import ChatMessage
+from core.domain.integration.integration_domain import IntegrationKind
 from core.domain.task_typology import TaskTypology
 from core.domain.types import AgentInput, AgentOutput
 from core.storage import TaskTuple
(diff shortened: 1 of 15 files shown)


api/core/domain/task_variant.py
```diff
@@ -40,6 +41,7 @@ class SerializableTaskVariant(BaseModel):
     is_public: bool | None = None
 
     creation_chat_messages: list[ChatMessage] | None = None
+    used_integration_kind: IntegrationKind | None = None
```
guillaq, line 44: You don't fill that value correct ?

yannbu: I do:
`# Register the integration used for the agent, so we can provide pertinent answers in the playground chat
                await self.storage.task_variants.update_task(
                    relevant_run_and_agent.agent.task_id,
                    used_integration_kind=integration_slug,
                )`

PR #301, opened 2025-05-20
[WOR-4696] [WOR-4686] [WOR-4667] [WOR-4611] [WOR-4693] [WOR-4636] [WOR-4807] [WOR-4755] [WOR-4828] Proxy Playgrounds - Refactor

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4696/proxy-playground-ui-polishes
https://linear.app/workflowai/issue/WOR-4686/allow-users-to-save-versions-that-dont-have-input-variables
https://linear.app/workflowai/issue/WOR-4667/update-input-variables-on-templated-message-changes
https://linear.app/workflowai/issue/WOR-4611/task-detail-view-uiux-updates-proxy-and-non-proxy
https://linear.app/workflowai/issue/WOR-4693/header-missing-on-edit-schema-modal-in-preview
https://linear.app/workflowai/issue/WOR-4636/proxy-code-page-ui-updates
https://linear.app/workflowai/issue/WOR-4807/automatically-moving-input-messages-to-version-messages
https://linear.app/workflowai/issue/WOR-4755/improved-code-accessibility-on-playground
https://linear.app/workflowai/issue/WOR-4828/new-code-blocks-for-markdown-chat-messages

Recordings:


anyacherniss:
**First round of QA done** 
Still want to regression test the regular playground; will do that tomorrow

- [ ] Add message button too far indented. Left side should be aligned with the left side of the messages

https://github.com/user-attachments/assets/8416bb40-f18c-471e-bc0d-8bbcbe8e35bd

CleanShot 2025-05-20 at 19 07 10@2x


- [ ] [not a blocker] maybe the edit button can show on top of the text to avoid the text jumping when hovered on?

https://github.com/user-attachments/assets/1d87498e-c9d4-4f51-b240-8f2fdcae1b47

- [ ] When adding a message in above/below a message, can the message type be the "opposite" of the original message that was hovered over to add the new message? For example: if the original message was a user message, the new message type would be an assistant message, if the original message was an assistant message the new message would be a user message.

(If adding a message above/below a system message or tool call, can default to user message)

https://github.com/user-attachments/assets/916dfa2e-eebf-49c3-8022-6df4d9f9accb

- [ ] When the view is too narrow, the deploy button can't be clicked. Maybe we can wrap the button on to a new line? (Open to other ideas, not sure of the best way to solve!)

https://github.com/user-attachments/assets/bb17916a-d0d2-4216-8a51-a1798659733a

- [ ] Not sure what's meant to be displayed here, but I think the current display is a bug?
CleanShot 2025-05-20 at 19 08 55@2x

- [ ] Version preview also show "view code"? Tapping view code would automatically save the version and then take the user to the code page.

CleanShot 2025-05-20 at 19 10 15@2x

- [ ] User should be able to switch between message types for the reply. For example an agent can respond with both a tool call and a message, and the user might want to respond to the message, not the tool call request.

CleanShot 2025-05-20 at 19 12 05@2x

- [ ] If there are no messages, I think the "Add message" button should always be shown (instead of just on hover)

https://github.com/user-attachments/assets/d2de090f-512d-430b-8238-e5bb4c0af0d6


- [ ] Since we aren't allowing the user to delete the list system message on the right, can the "x" be hidden or look disabled (ie. a lighter color)? Otherwise it looks like the button is broken

(see above screen recording)

- [ ] Missing updated tools UI 
CleanShot 2025-05-20 at 19 23 36@2x

CleanShot 2025-05-20 at 19 23 50@2x

CleanShot 2025-05-20 at 19 24 14@2x

CleanShot 2025-05-20 at 19 24 27@2x

- [ ] Variable formatting in the tooltip doesn't match the actually formatting, my bad! 🤦‍♀️ Should be `{{variable_here}}` (no spaces)
CleanShot 2025-05-20 at 19 27 12@2x

- [ ] Save button in versions header missing preview on hover
CleanShot 2025-05-20 at 19 35 20@2x

CleanShot 2025-05-20 at 19 35 43@2x

guillaq (approved):
Did not get a chance to do a full review but code looks ok

anyacherniss:
**QA is done**

✅ above issues all retested

- [ ] Probably a better UX to make the tool call result immediately editable (when it's being added as a response), instead of requiring the user to click edit after already clicking "Reply with tool call"


https://github.com/user-attachments/assets/7b96d5c6-52d0-4486-90bd-2ee3b3b0544b

- [ ] [not a blocker] right border of `creative` label isn't shown
CleanShot 2025-05-21 at 13 24 01@2x

- [ ] It looks like I can only switch between reply message types if the assistant message contains a tool call request. I'm okay to leave as is for now until we have a specific use case where the user would want a different reply type, if it's a lot of work. But I think ideally we always offer the option, for consistency.


https://github.com/user-attachments/assets/25f43c0e-616f-45fe-b239-3ace3e88be21

---

> This issue will take time and requires additional logic so let's move to another issue:

Sounds good.

anyacherniss:
**Another QA round done!**
linking my previous QA comments so they don't get buried (totally okay that the focus is on adding the high priority features first)
- https://github.com/WorkflowAI/WorkflowAI/pull/301#issuecomment-2898765760
- https://github.com/WorkflowAI/WorkflowAI/pull/301#issuecomment-2898797635

I'll open separate issues for UX items discovered/discussed outside of this QA!


- [ ] [NOT a blocker] Button gets cut off at the top Copy should also be "Load from..." (instead of "Try from..")
CleanShot 2025-05-28 at 17 13 47@2x
- [ ] [NOT a blocker] Update tooltip copy should be "Loads all messages and version"
CleanShot 2025-05-28 at 17 14 24@2x
- [ ] [NOT a blocker] clicking on the version variables shows a text cursor

https://github.com/user-attachments/assets/efac371a-a506-4e82-b7ab-161394916c04

- [ ] [Not a blocker, but should add as soon as reasonable], allow the user to copy all messages (user, assistant, system)

https://github.com/user-attachments/assets/9116ef41-15b2-445b-ae67-5bbd9b3580f5

- [ ] [not a blocker] @jacekzimonski might be a misunderstadning on my part, but I created a chat only using one agent and only see metadata for the last run on the detail view. Was hoping to see for all the assistant messages. 

Run: https://workflowai-fj63kzted-workflowai.vercel.app/workflowai/agents/flight-extractor-test/3/runs?page=0&taskRunId=019718c9-74f9-70cd-1172-ebce19955096


https://github.com/user-attachments/assets/63e88c46-5ccc-43b8-95c8-43e1bde40ad6

- [ ] [Not a blocker] On the current playground, when the temp is adjusted we run automatically. Can that behavior be added to the proxy?

https://github.com/user-attachments/assets/44ff5d79-b27c-4d69-9dfa-b80377c17024

- [ ] Model cell needs padding on top and below when the view is narrow

CleanShot 2025-05-28 at 17 36 53@2x


---

✅ passes QA: https://linear.app/workflowai/issue/WOR-4693/header-missing-on-edit-schema-modal-in-preview
✅ passes QA: user feedback (requested on slack)

anyacherniss:
Okay, I've created a new issue in linear that contains the unfixed bugs from this PR, so it can be merged. 

For the future, please try to avoid letting unresolved QA pile up OR if you foresee this happening tell me sooner so I can write QA feedback in linear directly to avoid needing to spend additional time duplicating my comments. As is, the visuals aren't loading in linear which isn't ideal, but it would take too long to download and reupload all images/videos form the original comments.

PR #303, opened 2025-05-21
Endpoint to generated tool call result preview

yannbu:
Closes [WOR-4721: Suggest tool call replies](https://linear.app/workflowai/issue/WOR-4721/suggest-tool-call-replies)

### DIFF
--- a/api/api/routers/runs_v1.py
+++ b/api/api/routers/runs_v1.py
@@ -3,14 +3,22 @@ from logging import getLogger
 from typing import Annotated, Any, Literal
 
 from fastapi import APIRouter, Query
+from fastapi.responses import StreamingResponse
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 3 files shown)


api/api/routers/runs_v1.py
```diff
@@ -261,3 +270,30 @@ async def save_run_version(
     version_id = (await runs_service.run_by_id(task_tuple, run_id, include={"version_id"})).group.id
     grp = await versions_service.save_version(task_tuple[0], version_id)
     return CreateVersionResponse.from_domain(grp)
+
+
+@router.post(
+    "/{run_id}/tool-call-result-preview",
```
guillaq, line 276:
```suggestion
    "/{run_id}/tools/previews/result",
```
?

guillaq (approved):

PR #321, opened 2025-05-27
[WOR-4767] Improve prompt textfield was removed and we are now showing error when the version was not created on the Model Output

jacekzimonski:
Frontend part of fixes.
Changes include:
- [x] remove direct improve prompt box @jacek
- [x] display a proper error message when template is invalid. @jacek POST /versions returns a 400 @jacek

@guillaq 
For now it's targeting Main, after the release branch will be created will change to the release branch


guillaq (commented):
Looks good, can you rebase to https://github.com/WorkflowAI/WorkflowAI/pull/324 ?

anyacherniss:
@jacekzimonski not sure if I'm doing something wrong, but I'm trying to test for the error message and not seeing one. 
I tried adding (manually, because I could not trigger improve prompt to do it) an input variable that doesn't exist, but the run completed. 

Let me know if I'm testing incorrectly!


https://github.com/user-attachments/assets/cc830117-49e6-4131-9819-8f7766f34b29

PR #347, opened 2025-06-02
[WOR-4781] Add data-1p-ignore to input fields

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4781/add-data-1p-ignore-to-input-fields


guillaq (approved):
So weird.

guillaq:
I think Xavier has a weird browser config. I have never had an issue with WAI showing 1p content

PR #348, opened 2025-06-02
Dedicated endpoint to improve system messages

yannbu:
Closes [WOR-4833: Dedicated endpoint to "improve system messages"](https://linear.app/workflowai/issue/WOR-4833/dedicated-endpoint-to-improve-system-messages)

### DIFF
--- a/api/core/agents/improve_version_messages_agent.py
+++ b/api/core/agents/improve_version_messages_agent.py
@@ -1,0 +1,88 @@
+import json
+import os
+from typing import Any, AsyncIterator, Literal, NamedTuple
+
+import workflowai
+from openai import AsyncOpenAI
+from pydantic import BaseModel
+
+
+class ImproveVersionMessagesInput(NamedTuple):
+    initial_messages: list[dict[str, Any]]
+
+    class Run(BaseModel):
+        input: str
+        output: str
+
+    run: Run | None
+    improvement_instructions: str | None
+
+
+class ImproveVersionMessagesOutput(BaseModel):
+    class Message(BaseModel):
+        role: Literal["system", "user", "assistant"]
+        content: str
+
+    improved_messages: list[Message]
+
+
+class ImproveVersionMessagesResponse(BaseModel):
+    improved_messages: list[dict[str, Any]]
+    feedback_token: str | None
+
+
+async def improve_version_messages_agent(
+    input: ImproveVersionMessagesInput,
+) -> AsyncIterator[ImproveVersionMessagesResponse]:
+    client = AsyncOpenAI(
+        api_key=os.environ["WORKFLOWAI_API_KEY"],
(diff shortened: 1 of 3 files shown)


api/core/agents/improve_version_messages_agent.py
```diff
@@ -0,0 +1,88 @@
...
+            feedback_token: str | None = getattr(chunk, "feedback_token", None)
+            if hasattr(chunk, "type") and chunk.type == "content.delta" and hasattr(chunk, "delta"):
+                if chunk.parsed and isinstance(chunk.parsed, dict):
+                    yield ImproveVersionMessagesResponse(
+                        **(chunk.parsed),  # pyright: ignore[reportUnknownMemberType, reportUnknownArgumentType]
+                        feedback_token=feedback_token,
+                    )
+                    continue
+
+                # It seems the runs from cache do not have a 'parsed' attribute filled,  so we need to parse the delta manually
```
yannbu, line 79: @guillaq FYI, not expected right ?

guillaq: The parsed attribute comes from the openai sdk not from their API, no ?

guillaq (commented):
The OpenAI SDK only streams delta I believe. Not full python objects ?
If we need full python objects we could add an option to the "stream options" to pass full json objects to make it easier

api/core/agents/improve_version_messages_agent.py
```diff
@@ -0,0 +1,88 @@
...
+                if chunk.delta:
+                    try:
+                        agg += chunk.delta
+                        yield ImproveVersionMessagesResponse(
+                            **(json.loads(agg)),
```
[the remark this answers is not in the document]
guillaq: Super weird, i'll add a test to autopilot

guillaq (approved):

api/core/agents/improve_version_messages_agent.py
```diff
@@ -0,0 +1,88 @@
...
+        },
+        temperature=0.0,
+    ) as response:
+        chunk = None
+        agg = ""
+        async for chunk in response:
+            feedback_token: str | None = getattr(chunk, "feedback_token", None)
+            if hasattr(chunk, "type") and chunk.type == "content.delta" and hasattr(chunk, "delta"):
+                if chunk.parsed and isinstance(chunk.parsed, dict):
+                    yield ImproveVersionMessagesResponse(
```
guillaq, line 73: that seems very dangerous ? maybe try / catch

yannbu:
@guillaq I'll merge like in this, in order to overrid preview, I'll address your comments later.

api/core/agents/improve_version_messages_agent.py
yannbu: yes, yes, but still the behaviour may be caused by cases when the whole stream is a single sse ?

PR #356, opened 2025-06-03
[WOR-4856] Update generated code on code page when deployment changes

jacekzimonski:
- Update Generated Code on Code Page when Deployment changes
- We should change the behavior of showing the banner for deployment to: Show only if the version is not deployed to any env. Current that we are changing: show it always if it's not deployed to all envs.
- Show entry for the version deployed to env only once on Proxy, currently we are showing one for env and one more copy for not deployed to env

+ to make that work I made the quick change between version work! (video in demo section)

**Demos:**
No refreash Quick Change
Deploy refreash


guillaq (approved):

anyacherniss:
This passes QA 👍

PR #357, opened 2025-06-03
[WOR-4861] Grouping of Runs clean up on Frontend

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4861/grouping-of-runs-clean-up-on-frontend

Change List:
- We are not showing run details for messages in Playground, only in Run Details Modal
- We support completely new Preview For Run, on run page!
- This new preview shows:
- Input Variables if they are there. (For Proxy Only)
- Messages. (For Proxy Only)
- The Message order in the Run Details Modal and Preview for Run changed the order:
- We first show potential errors if they are there
- Then the newest message and etc.
- Fixed the issue with clipping the Open in Playground badge

Demo:
Screenshot 2025-06-03 at 2 19 48 PM
Screenshot 2025-06-03 at 2 19 54 PM
Screenshot 2025-06-03 at 2 20 20 PM
Screenshot 2025-06-03 at 2 20 28 PM


anyacherniss:
**QA is done**

I tested the expected behavior for this PR although I know we've since decided to undo the new message ordering added here

- [ ] [not a blocker] Should probably hide the "try in playground" button in the preview

https://github.com/user-attachments/assets/03af1dec-4523-4e5e-ba45-7d5df4e3150d

PR #367, opened 2025-06-04
[WOR-4814] Images as input variables UX

jacekzimonski:
@guillaq @anyacherniss 
PR Ready!


guillaq (approved):

anyacherniss:
**QA is done**

- [ ] Not sure if backend or frontend, but when switching to audio, I wasn't able to select an audio model until after I ran the task once on a model I knew wouldn't work. 

Will probably be tricky to fix since the schema with audio isn't officially created until that first run is complete, so it can be a follow up if needed, but not a great UX as is.


CleanShot 2025-06-09 at 12 30 10@2x

guillaq:
Basically you want an endpoint that returns models based on the schema ? We should do as a follow up. We indeed need a new endpoint.
-> https://linear.app/workflowai/issue/WOR-4918/endpoint-to-get-models-based-on-a-given-io-schema

anyacherniss:
Yes all good to merge and fix parts requiring the new endpoint later! 👍

PR #388, opened 2025-06-07
Fix object check in ProxyTools

pierrevalade:
## Summary
- fix runtime error when tool is not an object in ProxyTools

## Testing
- `make lint` *(fails: Import "taskiq" could not be resolved)*
- `yarn --cwd client test`

------
https://chatgpt.com/codex/tasks/task_e_6844a128a9a483219c46c73826c3f1a9


pierrevalade:
@jacekzimonski this issue is a bug that should be merged directly by you, i think, why do you need  @anya?

guillaq:
@pierrevalade do you have a URL that triggered the issue ? It would be interesting to see why Tool_Output was not an object

anyacherniss:
> The current procedure was to QA every frontend PR we have.
@anyacherniss @guillaq to confirm

As a general practice yes. That said, since this looks like a small, one line change, if we're confident that it won't introduce regressions, I'm fine if it is merged to directly if it takes more time than it's worth to provide context for me to manually QA.

guillaq:
It's in preview so we basically can't QA in the preview branch. The code looks sound. We can just merge and do some QA when it's in preview.

PR #386, opened 2025-06-06
feat: Add MCP (Model Context Protocol) POC specification for Workflow…

pierrevalade:
…AI integration

PR #403, opened 2025-06-10
MCP server POC

yannbu:
Related to https://linear.app/workflowai/issue/WOR-4911/workflowai-mcp-v01


mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+        async with httpx.AsyncClient(timeout=60.0) as client:
+            response = await client.post(
+                f"{WORKFLOWAI_API_URL}/{TENANT}/agents/{agent_id}/prompt-engineer-agent/messages",
+                json=body,
+                headers=headers,
+            )
+            response.raise_for_status()
+
+            last_messages: List[Dict[str, Any]] = []
```
pierrevalade, line 96: is that because `prompt-engineer-agent/messages` endpoint only support streaming?

mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+        logger.error("Error calling meta agent API", extra={"error": str(e)})
+        return {
+            "success": False,
+            "error": "Failed to call AI engineer API",
+            "details": str(e),
+        }
+
+
+@mcp.tool
+async def fetch_getting_started_docs() -> dict[str, Any]:
```
pierrevalade, line 132: @yannbu think early about how to support multiple programming languages, a developer writing code in Typescript, Rust, etc... (no everyone will be in Python)

search for the term `programming` in this PR https://github.com/WorkflowAI/WorkflowAI/pull/386/files#diff-aac09eb14a7fb943fc66ebf9138fbcad4f1e5a5d8a6ea71fc83bd2088a806f64R78 and you'll see what I started.

pierrevalade:
> Still WIP, th code will need to be integrated with the backend code. I'm using the API for now, as my priority was to go fast and test the interaction between the WorkflowAI MCP and Cursor.

why not continue to use the API actually? could you please explain? thanks.

mcp-server/main.py
yannbu: yes

mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+
+def _headers() -> dict[str, str]:
+    return {
+        "Accept": "application/json",
+        "Authorization": f"Bearer {FIXED_WORKFLOWAI_API_KEY}",
+    }
+
+
+@mcp.tool
+async def get_ai_engineer_response(
```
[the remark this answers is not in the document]
guillaq: Do we think that cursor will actually do a back and forth or even preserve the conversation_id between calls ? 
I would think that it would be an early optim at this stage ?

mcp-server/main.py
guillaq: an example reason why we should not wrap the API

mcp-server/main.py
guillaq: This should be a full text search in the docs instead ?

mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+
+    return {
+        "success": True,
+        "documentation": docs,
+    }
+
+
+@mcp.tool
+async def list_available_models() -> dict[str, Any]:
```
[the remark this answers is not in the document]
guillaq: passing a "typology" (input image, etc.) might be more flexible. For example cursor might want to get the list of models BEFORE the agents has been created or ran.

mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+        logger.error("Error calling models API", extra={"error": str(e)})
+        return {
+            "success": False,
+            "error": "Failed to call models API",
+            "details": str(e),
+        }
+
+
+@mcp.tool
+async def fetch_run_details(agent_id: str, run_id: str) -> dict[str, Any]:
```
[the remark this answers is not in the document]
guillaq: agreed on allowing agent_id, run_id and run_url. all optional.

Best way would be to add a way to fetch a run only by ID which should be possible.

mcp-server/main.py
guillaq: it should take `keywords` and `language` indeed. Maybe also an optional `agent_id` ?

mcp-server/main.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+        return {
+            "success": False,
+            "error": "Failed to call agents APIs",
+            "details": str(e),
+        }
+
+
+@mcp.tool
+async def list_agent_versions(
```
guillaq, line 376: How would cursor know the schema_id ? it's not in cursor if there are no deployments.

guillaq (commented):
> why not continue to use the API actually? could you please explain? thanks.

MCP tools != API. If we keep wrapping the API we will need to add dedicated endpoints, maintain synchronization, etc.

The code is sufficient layered. We should just be able to use the service layer with very little code duplication.

mcp-server/main.py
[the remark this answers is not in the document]
yannbu: OK, I'll implement that (leaving this comment open as a TODO)

mcp-server/main.py
yannbu: This is something I still need to figure out, also when calling the AI engineer. Cursor can pass an agent ID, but cursor can't know the schema ID (unless agent is deployed as you mentioned)

I need to think of ways the AI engineer can "guess" the agent schema id based on what Cursor will pass (ex: existing code block) OR think about how the AI engineer can work without a agent schema id.

Any ideas @guillaq ?

guillaq: >  need to think of ways the AI engineer can "guess" the agent schema id based on what Cursor will pass

I think that will be a bit complicated and error prone.

What flow are we trying to solve ?

mcp-server/main.py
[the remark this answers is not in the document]
pierrevalade: > Do we think that cursor will actually do a back and forth or even preserve the conversation_id between calls ?

yes, I tried in my DEMO and Cursor will pass the `reply_to_message_id`, when expected.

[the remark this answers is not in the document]
yannbu: Tracked here [WOR-4949: MCP • Implement back and forth discussion between Cursor and AI Engineer with 'reply_to_message_id'](https://linear.app/workflowai/issue/WOR-4949/mcp-implement-back-and-forth-discussion-between-cursor-and-ai-engineer)

I'll close this conversation to be able to merge the PR.

mcp-server/main.py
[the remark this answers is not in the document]
yannbu: Tracked [WOR-4949: MCP • Implement back and forth discussion between Cursor and AI Engineer with 'reply_to_message_id'](https://linear.app/workflowai/issue/WOR-4949/mcp-implement-back-and-forth-discussion-between-cursor-and-ai-engineer).

api/api/routers/agents/ai_engineer_router.py
```diff
@@ -0,0 +1,83 @@
+from typing import Annotated, AsyncIterator
+
```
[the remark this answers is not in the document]
yannbu: well spotted ! thanks https://github.com/WorkflowAI/WorkflowAI/pull/403/commits/cb175d4e0ee4c197c95a5bd3f74b94e1563fa6a3

PR #409, opened 2025-06-11
Add pricing exposure plan

pierrevalade:
## Summary
- add plan for exposing model pricing
- clarify API response and exclude image/audio pricing

## Testing
- `git status --short`

------
https://chatgpt.com/codex/tasks/task_e_684900d71db08321a0eb8772958893a4


pierrevalade:
@guillaq @yannbu I've tried this prompt here:
> I would like to expose the pricing of each models under the /v1/models endpoint.
> Could you please write a plan.MD to explain how you would approach this change?

the goal of the `plan.MD` is make sure the AI gets a change to put its thoughts together before moving forward with the actual implementation. (remember when we were doing this ourselves too? 😉 )

cc @guillaq to review the actual `PLAN.md` -- since Codex does not read, to my knowledge, GitHub comment, maybe you can update the plan directly or at least add suggestion in GitHub that we can commit directly and we can ask Codex to do another pass. Thank you.

guillaq (commented):
Plan is sound

guillaq:
Plans are good but I think they might not be needed for scoped features.

See https://github.com/WorkflowAI/WorkflowAI/pull/411 that was created without a plan.

```
Can you add the price information (per token, input / output) to the /v1/models endpoint in main.py ? 
```

PR #410, opened 2025-06-11
Add agent_id field in stats endpoint

pierrevalade:
## Summary
- expose `agent_id` in AgentStat and mark `agent_uid` as deprecated
- adjust client types and store to use the new field
- add a plan outlining the deprecation steps

## Testing
- `poetry run ruff check .`
- `poetry run pyright .`
- `poetry run pytest api/tests/component/authentication/authentication_test.py::test_unknown_user -q` *(fails: ServerSelectionTimeoutError)*
- `yarn prettier-check`
- `yarn workspace workflowai lint`
- `yarn workspace workflowai build` *(fails: JavaScript heap out of memory)*

------
https://chatgpt.com/codex/tasks/task_e_68490b3c318c83218944a869b958add7


api/api/routers/agents_v1.py
```diff
@@ -156,7 +157,12 @@ async def get_agent_stats(
 ) -> Page[AgentStat]:
     from_date = from_date or datetime.now() - timedelta(days=7)
     items = [
-        AgentStat(agent_uid=stat.agent_uid, run_count=stat.run_count, total_cost_usd=stat.total_cost_usd)
+        AgentStat(
+            agent_id=str(stat.agent_uid),
```
guillaq, line 161: are you looking for the string agent_id as in `meta-agent` for example ? What you have here would give you a stringified int (`1234`)

guillaq:
Closing, more work is needed than just stringifying the UID.
Follow up in https://linear.app/workflowai/issue/WOR-5065/stop-exposing-agent-uid-only-expose-agent-id

PR #425, opened 2025-06-12
add initial discussion on FastAPI-MCP usage and concerns

pierrevalade:
@guillaq @yannbu


yannbu:
@pierrevalade I think you can close, I've applied your remarks in https://github.com/WorkflowAI/WorkflowAI/pull/438 and we stitched to FastMCP

PR #439, opened 2025-06-13
Use templates for code gen

yannbu:
From my work from last week.

Related to https://linear.app/workflowai/issue/WOR-4811/improvements-on-generated-code


pierrevalade:
@guillaq will work on this PR next week.

PR #450, opened 2025-06-14
Add ability to stream entire payloads

guillaq:
ref https://linear.app/workflowai/issue/WOR-4963/stream-full-objects-in-openai-proxy

### DIFF
--- a/api/api/routers/openai_proxy/_openai_proxy_models.py
+++ b/api/api/routers/openai_proxy/_openai_proxy_models.py
@@ -304,10 +304,12 @@ class OpenAIProxyResponseFormat(BaseModel):
 class OpenAIProxyStreamOptions(BaseModel):
     include_usage: bool | None = None
 
-    aggregate_content: bool | None = Field(
+    valid_json_chunks: bool | None = Field(
         default=None,
-        description="Whether to send the aggregate the content server side and stream the entire payload "
-        "(as opposed to sending the delta). When using a json based format, the content is always a valid JSON object.",
+        description="Whether to only send valid JSON chunks when streaming. When set to true, the delta aggregation "
+        "is performed by WorkflowAI and only valid aggregated JSON strings are sent in place of the content delta. "
+        "valid_json_chunks is only relevant when using a json based response format (either `json_object` or "
+        "`json_schema`).",
     )
 
     model_config = ConfigDict(extra="allow")
--- a/api/api/routers/openai_proxy/_openai_proxy_models.py
+++ b/api/api/routers/openai_proxy/_openai_proxy_models.py
@@ -304,6 +304,12 @@ class OpenAIProxyResponseFormat(BaseModel):
 class OpenAIProxyStreamOptions(BaseModel):
     include_usage: bool | None = None
 
+    aggregate_content: bool | None = Field(
+        default=None,
+        description="Whether to send the aggregate the content server side and stream the entire payload "
+        "(as opposed to sending the delta). When using a json based format, the content is always a valid JSON object.",
+    )
+
     model_config = ConfigDict(extra="allow")
 
 
@@ -885,12 +891,19 @@ class OpenAIProxyChatCompletionChunkDelta(BaseModel):
     role: Literal["user", "assistant", "system", "tool"] | None
 
     @classmethod
-    def from_domain(cls, output: RunOutput, deprecated_function: bool):
-        if not output.delta and not output.tool_call_requests:
+    def from_domain(
+        cls,
+        output: RunOutput,
+        output_mapper: Callable[[AgentOutput], str | None],
+        deprecated_function: bool,
+        aggregate_content: bool | None,
+    ):
+        if not aggregate_content and not output.delta and not output.tool_call_requests:
             return None
+
         return cls(
             role="assistant",
-            content=output.delta,
+            content=output_mapper(output.task_output) if aggregate_content else output.delta,
             function_call=OpenAIProxyFunctionCall.from_domain(output.tool_call_requests[0])
             if deprecated_function and output.tool_call_requests
             else None,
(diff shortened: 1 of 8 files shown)


api/api/routers/openai_proxy/_openai_proxy_models.py
```diff
@@ -304,6 +304,12 @@ class JsonSchema(BaseModel):
 class OpenAIProxyStreamOptions(BaseModel):
     include_usage: bool | None = None
 
+    aggregate_content: bool | None = Field(
```
guillaq, line 307: @pierrevalade only interesting part of the PR. `aggregate_content` changes the behavior of the stream to send aggregated data instead of deltas (same behavior as our run endpoint). This way @jacekzimonski and users that want to handle JSON can use the payloads as is.

```python
    streamer = await openai_client.chat.completions.create(
        model="gpt-4o",
        messages=[{"role": "user", "content": "Hello, world!"}],
        stream=True,
        stream_options=cast(ChatCompletionStreamOptionsParam, {"aggregate_content": True}),
    )
```

pierrevalade: @guillaq if your intention is to expose this API fields externally, could you please write the documentation part?

pierrevalade (changes requested):

api/api/routers/openai_proxy/_openai_proxy_models.py
guillaq: Added a section in the documentation. I think the `aggregate_content` feature is a very nice addition.

pierrevalade:
@guillaq please request another review when ready for me. thank you.

docsv2/content/docs/inference/streaming.mdx
```diff
@@ -3,12 +3,15 @@ title: Streaming
...
+import { Tab, Tabs } from 'fumadocs-ui/components/tabs';
+
 By default, when you make a request to the WorkflowAI API, the model generates the entire output before returning it in a single HTTP response. For longer outputs, this can mean waiting until the full response is ready. With streaming, you can start receiving and processing the model's output as it is generated, allowing you to display or use partial results in real time.
 
 ## Enable streaming
 
-<Callout type="success">
-  Streaming is available exactly like in the OpenAI API. If you are already streaming from the OpenAI API, no code change is required.
+<Callout type='success'>
+  Streaming is available exactly like in the OpenAI API. If you are already streaming from the OpenAI API, no code
```
pierrevalade, line 13: @guillaq do you know what changed here, except the new line?

docsv2/content/docs/inference/streaming.mdx
```diff
@@ -94,4 +100,94 @@ stream = client.chat.completions.create(
...
+```sh
+{"name
+":"
+"John
+Doe
+",
+...
+```
+
+By setting the `stream_options.aggregate_content` parameter to `true`, WorkflowAI can aggregate the deltas into valid partial JSON objects, transforming the above deltas into:
```
pierrevalade, line 130: I would suggest `valid_json_chunks` as a parameter name. 

### Problem-Solution Clarity
- **`aggregate_content`**: Vague - doesn't specify what content or why you'd aggregate it
- **`valid_json_chunks`**: Crystal clear - you want JSON chunks that are valid/parseable

### Domain Specificity
- **`aggregate_content`**: Generic - could apply to any content type (text, binary, etc.)
- **`valid_json_chunks`**: JSON-specific - immediately tells you this is for JSON streaming scenarios

### Developer Mental Model
- **`aggregate_content`**: Developer thinks "what's being aggregated and why?"
- **`valid_json_chunks`**: Developer thinks "ah, I'll get parseable JSON instead of fragments"

### Self-Documenting Code
- **Current**: `stream_options={"aggregate_content": True}` # What does this do?
- **Proposed**: `stream_options={"valid_json_chunks": True}` # Obviously gives valid JSON chunks

### Pain Point Alignment
- **`aggregate_content`**: Describes the technical mechanism (aggregation)
- **`valid_json_chunks`**: Describes the desired outcome (valid, parseable chunks)

### Discoverability
- **`aggregate_content`**: Developers won't search for "aggregate" when they want parseable JSON
- **`valid_json_chunks`**: Matches what developers actually search for: "valid JSON streaming"

### Future-Proof Naming
- **`aggregate_content`**: If the internal implementation changes, the name becomes misleading
- **`valid_json_chunks`**: Describes the end result regardless of how it's achieved internally

**Bottom line**: `valid_json_chunks` directly addresses the developer's need ("I want JSON I can parse") while `aggregate_content` focuses on implementation details they don't care about.

api/api/routers/openai_proxy/_openai_proxy_models.py
```diff
@@ -304,6 +304,12 @@ class JsonSchema(BaseModel):
...
+        default=None,
+        description="Whether to send the aggregate the content server side and stream the entire payload "
```
pierrevalade, line 309: the setting is only relevant when using some specific `response_format` right? should be clarified.

pierrevalade (changes requested):
thank you. ready for another pass.

docsv2/content/docs/inference/streaming.mdx
guillaq: The content was within a code block. So prettier formatted the content. Changed the default formatter for mdx files to be the mdx extension. We will lose auto code formatting in code blocks but the long lines should no longer be affected.

Although please notes that long lines are usually not a good idea when using git, since a change in a single word will mark a change in the entire line, as opposed to isolating the change. That's why MD and MDX mostly ignore new lines.

docsv2/content/docs/inference/streaming.mdx
guillaq: 👍

api/api/routers/openai_proxy/_openai_proxy_models.py
guillaq: Field documentation changed to

>Whether to only send valid JSON chunks when streaming. When set to true, the delta aggregation "
>"is performed by WorkflowAI and only valid JSON strings are sent in place of the content delta. "
>"valid_json_chunks is only relevant when using a json based response format (either `json_object` or "
>"`json_schema`

guillaq:
Renamed valid_json_chunks in docs and tests

docsv2/content/docs/inference/streaming.mdx
```diff
@@ -94,4 +99,94 @@ stream = client.chat.completions.create(
...
+
+```sh
+# Chunks that do not represent an update of the JSON object are ignored
+# Each chunk is a valid JSON object
+{"name": "John"}
+{"name": "John Doe"}
+...
+```
+
+<Tabs items={["Python", "TypeScript", "JSON"]}>
```
pierrevalade, line 139: I think the last option should be adjusted to "curl" ?

PR #459, opened 2025-06-16
Plug docs v2 to agent

yannbu:
For the playground agents and MCP AI Engineer

### DIFF
--- a/api.Dockerfile
+++ b/api.Dockerfile
@@ -61,7 +61,7 @@ RUN pip install --no-cache /wheels/*
 
 # Copy the rest of the application
 COPY api/api api
-COPY docs docs
+COPY docsv2 docsv2
 COPY api/core core
 COPY api/start.sh /app/start.sh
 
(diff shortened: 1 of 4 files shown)


pierrevalade:
I love that using a AI agent for RAG also means that we (our team at WorkflowAI) get observability built-in on the "RAG" via WorkflowAI, similar to what we discussed about MCP observability on top of a AI agent during today's checkin. 

in a world where inference becomes so cheap, sending data to a AI agent that stores all the runs automatically might be the new persistent storage.

pierrevalade:
> Is there any reason to still include `docs` in the repo then ?

maybe the playground running on non-proxy agents?

guillaq:
> maybe the playground running on non-proxy agents?

I don't think so ? The `docs` directory is no longer included in the dockerfile so it will be completely ignored by the API I believe

yannbu:
thanks guys, I'll merge this and keep docs v1 for the legacy playground agent. I need to wire up the legacy agent to the docsv1, in my PR for today.

PR #460, opened 2025-06-16
Verify agent prefix documentation accuracy

pierrevalade:
Documentation was updated to align with the clarified preferred method for agent identification.

Changes include:
*   Examples in `evaluations/user-feedback.mdx`, `inference/caching.mdx`, `inference/cost.mdx`, `inference/reliability.mdx`, and `reference/prompt-templating.mdx` were modified.
    *   These now use `metadata={"agent_id": "..."}` instead of an agent prefix within the `model` parameter.
*   `inference/models.mdx` was updated to explicitly state that `agent_id` in the `metadata` field is the preferred identification method.
    *   It also acknowledges the agent ID as a prefix in the `model` parameter as an acceptable alternative when metadata editing is not possible.
*   A callout was added to `quickstarts/openai-agents.mdx` to clarify that it demonstrates the alternative model parameter prefix method.
*   Changes previously made to `docsv2/content/docs/observability/index.mdx` were reverted, restoring the file to its original state.

The updates ensure documentation consistently promotes the preferred `agent_id` in metadata approach, while retaining and clarifying the alternative model parameter prefix method where contextually appropriate.


pierrevalade:
> @cursoragent I see more accurence of the agent id in model string, for example in mcp.mdx; is that on purpose ?

not working yet.

pierrevalade:
I will follow up with CursorAI manually now.

pierrevalade:
@yannbu should be better now.

PR #461, opened 2025-06-16
[WOR-4751] playground version, 2 small feedbacks

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-4751/playground-version-2-small-feedbacks

Demo:
Screen Recording 2025-06-16 at 4 47 11 PM


anyacherniss:
**QA is done** 

- [ ] When using the "Add message" button, the version message doesn't default to User message when expected. Currently only works if the message is added via the "+" button


https://github.com/user-attachments/assets/c92f4a91-f051-45cc-9459-e7ea9b1e1a32

anyacherniss:
This passes QA 👍

PR #484, opened 2025-06-17
Primitive: Evaluations

guillaq:
cf https://github.com/WorkflowAI/fumadocs-demo/pull/3

### DIFF
--- a/docsv2/content/docs/evaluations/meta.json
+++ b/docsv2/content/docs/evaluations/meta.json
@@ -3,6 +3,7 @@
   "icon": "ClipboardCheck",
   "pages": [
     "index",
+    "setup",
     "side-by-side",
     "user-feedback"
   ]
--- a/docsv2/content/docs/evaluations/setup.mdx
+++ b/docsv2/content/docs/evaluations/setup.mdx
@@ -1,0 +1,723 @@
+---
+title: Setup
+description: Automatically evaluate your AI agent outputs for quality, accuracy, and custom metrics.
+---
+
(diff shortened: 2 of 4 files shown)


guillaq:
@pierrevalade @yannbu  to follow up the discussions from https://github.com/WorkflowAI/fumadocs-demo/pull/3 I think I am still a bit confused.

If I understand correctly, the process described in this PR is:
- client triggers a run
- client fetches the output and evaluates the output (using an other agent or not)
- client POSTs results of the evaluation to WorkflowAI and WAI displays the results in benchmarks, etc.

In that context I don't really understand some parts of https://github.com/WorkflowAI/fumadocs-demo/pull/3#issuecomment-2964035395

> sorry, I mean the runs that are being evaluated are happening client-side.

The evaluation itself is also executed client side no ?

> yes. evaluations is an array.

If the evaluation itself is executed client side, why does the backend need to know about the evaluation configuration ? Would a "version of the evaluator" not suffice so that the backend knows how to dismiss stale reviews ?

> I'm OK to let the end-user run the evaluator agent directly and post a /evaluations directly. the idea of attaching the evaluators: [] isn't maybe as valuable as the idea to configure the evaluator through code.

I think that would be clearer


To be clear, I think we could separate the responsabilities between client and WAI:

Client is responsible for
- building an evaluation dataset
- creating versions
- running evaluations for given runs

WAI is responsible for:
- making sure there is 1 run per version per input in the evaluation dataset (no need for that to happen client side IMO, since WAI needs the information to correctly display benchmarks or evaluation results)
- receiving and displaying evaluation results

PR #486, opened 2025-06-17
More concise list models tool result

yannbu:
related to https://linear.app/workflowai/issue/WOR-4988/limite-size-of-available-model-list-currently-27k-tokens

BEFORE
26112 tokens

AFTER
11259 tokens

@pierrevalade @guillaq do you think we need to go deeper in the optimisation ?

### DIFF
--- a/api/api/routers/mcp/_mcp_models.py
+++ b/api/api/routers/mcp/_mcp_models.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel
 
 from api.schemas.user_identifier import UserIdentifier
 from api.schemas.version_properties import ShortVersionProperties
+from api.services.models import ModelForTask
 from core.domain.message import Message
 from core.domain.models.model_data import FinalModelData
 from core.domain.models.model_data_supports import ModelDataSupports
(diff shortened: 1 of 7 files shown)


api/api/routers/mcp/_mcp_models.py
```diff
@@ -16,6 +17,55 @@
...
 
 
+class ConciseLatestModelResponse(BaseModel):
+    id: str
+    currently_points_to: str
+
+
+class ConciseModelResponse(BaseModel):
+    id: str
+    maker: str
```
guillaq, line 27: That's not the maker really ? Do we need that data ?

api/api/routers/mcp/_mcp_models.py
```diff
@@ -16,6 +17,55 @@
...
+    @classmethod
+    def from_model_data(cls, id: str, model: FinalModelData):
+        IGNORE_SUPPORTS = {"structured_output", "support_system_messages", "json_mode"}
+
+        provider_data = model.providers[0][1]
+        return cls(
+            id=id,
+            maker=model.provider_name,
+            display_name=model.display_name,
+            supports=[
```
guillaq, line 44: is that going to look weird when serialized in JSON ? we should maybe serialize to strings ?

guillaq (approved):

api/api/routers/mcp/_mcp_models.py
yannbu: @guillaq that's just a list of strings:


 ```
 "supports": [
          "input_image",
          "output_text",
          "json_mode",
          "structured_output",
          "parallel_tool_calls",
          "tool_calling"
        ],
```

File: api/api/routers/mcp/_mcp_models.py (hunk 1 above)
Comment (yannbu) [reply api/api/routers/mcp/_mcp_models.py:27]: should I simply say "provider" ?

I think that the info is useful.

Comment (yannbu) [reply api/api/routers/mcp/_mcp_models.py:27]: Cursor might not know right away which provider build which models (especially new models)

File: api/api/routers/mcp/_mcp_models.py [hunk 3]
```diff
@@ -16,6 +17,55 @@
...
+    currently_points_to: str
+
+
+class ConciseModelResponse(BaseModel):
+    id: str
+    maker: str
+    display_name: str
+    supports: list[str]
+    quality_index: int
+    cost_per_input_token_usd: float
```
pierrevalade, line 31: is the cost per 1 input token or per 1k tokens?

yannbu:
@guillaq thanks for the review !
I'll merge to keep the ball rolling, but feel free to answer my messages in conversations.

api/api/routers/mcp/_mcp_models.py
```diff
@@ -16,6 +17,55 @@
...
+    cost_per_output_token_usd: float
+    release_date: str
+
+    @classmethod
+    def from_model_data(cls, id: str, model: FinalModelData):
+        IGNORE_SUPPORTS = {"structured_output", "support_system_messages", "json_mode"}
```
pierrevalade, line 37: I think the opposite should be implemented, a "allow" list versus a ignore, to avoid new fields being added here without anyone noticing.

yannbu:
> > @pierrevalade @guillaq do you think we need to go deeper in the optimisation ?
> 
> @yannbu what % of tokens would be saved by removing `supports` completely?

@pierrevalade like 20% more saving (around 2k tokens)

PR #488, opened 2025-06-17
Gemini 2.5 updates

pierrevalade:
To adjust to the Gemini 2.5 thinking model updates, several changes were implemented:

*   New stable model enums, `GEMINI_2_5_PRO`, `GEMINI_2_5_FLASH`, and `GEMINI_2_5_FLASH_LITE`, were added to `api/core/domain/models/models.py`.
*   Pricing data in `api/core/domain/models/model_provider_datas_mapping.py` was updated for both Google Vertex AI and Gemini API providers.
    *   Unified pricing for `Flash` models was introduced, removing the distinction between thinking and non-thinking costs.
    *   Pricing for the new `Flash-Lite` model was added.
    *   Existing preview `Flash` model pricing was adjusted to align with the new unified structure.
*   A new documentation file, `GEMINI_2_5_UPDATES.md`, was created. This file details the changes from the Google blog post, including new model availability, updated pricing, deprecation timelines, and implementation status, providing migration guidance.

These changes ensure compatibility with the latest Gemini 2.5 models and their updated cost structures, offering cost optimization and simplified pricing.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,7 +14,7 @@
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
   "python.testing.pytestArgs": ["-vv", "-s"],
-  "python.languageServer": "Pylance",
+  "python.languageServer": "None",
   "python.analysis.autoImportCompletions": true,
   "editor.suggestSelection": "first",
   "editor.quickSuggestions": {
--- a/api/core/domain/models/model_provider_datas_mapping.py
+++ b/api/core/domain/models/model_provider_datas_mapping.py
@@ -21,29 +21,7 @@ ONE_MILLION_TH = 0.000_001
 
 
 GOOGLE_PROVIDER_DATA: ProviderDataByModel = {
-    Model.GEMINI_2_5_FLASH_THINKING_PREVIEW_0520: ModelProviderData(
-        text_price=TextPricePerToken(
-            prompt_cost_per_token=0.15 * ONE_MILLION_TH,
-            completion_cost_per_token=3.50 * ONE_MILLION_TH,
-            source="https://ai.google.dev/gemini-api/docs/pricing#gemini-2.5-flash",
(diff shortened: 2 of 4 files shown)


pierrevalade:
@guillaq 100% AI-generated from this prompt:
> open a PR to adjust to the changes from https://developers.googleblog.com/en/gemini-2-5-thinking-model-updates/

so expecting a bit of mess up to clean, and use as a an opportunity to write some instructions on how to add/edit models?

api/core/domain/models/model_provider_datas_mapping.py
```diff
@@ -21,41 +21,63 @@
 
 
 GOOGLE_PROVIDER_DATA: ProviderDataByModel = {
-    Model.GEMINI_2_5_FLASH_THINKING_PREVIEW_0520: ModelProviderData(
```
pierrevalade, line 24: instead of removing the model `Model.GEMINI_2_5_FLASH_THINKING_PREVIEW_0520` -- I assume the AI should have marked as deprecated? (but probably does not know about the "process" we have).

.vscode/settings.json
```diff
@@ -14,7 +14,7 @@
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
   "python.testing.pytestArgs": ["-vv", "-s"],
-  "python.languageServer": "Pylance",
+  "python.languageServer": "None",
```
guillaq, line 17: that's a bit strange. it would break all our Cursor setups

api/core/domain/models/model_provider_datas_mapping.py
guillaq: updating here would be ok but it missed the fact that it needed to update another file. I'll add the information to CONTRIBUTING.md since it's for both AGENTs and humans. (AGENTS.md points to CONTRIBUTING.md)

PR #492, opened 2025-06-18
Gemini 2.5 GA

guillaq:
- add gemini 2.5 Pro, Flash and Flash lite and update the pricing
- deprecate GPT 4.5 Preview


CONTRIBUTING.md
```diff
@@ -36,3 +36,58 @@ trigger deployments to the production environment. Once everything is ok, the br
 A hotfix allows fixing bugs in production without having to push changes to the development environment first.
 A hotfix branch should be created from the latest tag and a PR targeting main should be created. The flow is then the
 same as the release process.
+
+### Adding new models
```
yannbu, line 40: thanks!

api/core/domain/models/model_datas_mapping.py
```diff
@@ -729,41 +709,41 @@ def _raw_model_data() -> dict[Model, ModelData | LatestModel | DeprecatedModel]:
...
             provider_name=DisplayedProvider.GOOGLE.value,
             supports_tool_calling=True,
-            reasoning_level="none",
             fallback=ModelFallback.default(
                 "cheapest",
                 content_moderation=Model.GPT_41_NANO_LATEST,
                 rate_limit=Model.GEMINI_2_0_FLASH_001,
                 context_exceeded="no",
             ),
+            is_default=True,
```
pierrevalade, line 741: @guillaq would be helpful as a backup that updating `is_default` requires updating a unit-test as well, to double check that the list of default models is explicitly the list we intend?

guillaq: Yes, we could basically test that `is_default` is correctly set for the first 3 models in the enum. Which would have failed in my case.

api/core/domain/models/model_datas_mapping.py
```diff
@@ -802,27 +779,28 @@ def _raw_model_data() -> dict[Model, ModelData | LatestModel | DeprecatedModel]:
...
+                source="https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash-lite",
             ),
             icon_url="https://workflowai.blob.core.windows.net/workflowai-public/google.svg",
-            release_date=date(2025, 6, 6),
+            release_date=date(2025, 6, 18),
             quality_data=QualityData(
-                gpqa_diamond=86.4,
-                mmlu=89.2,
-                source="https://deepmind.google/models/gemini/pro/",
+                gpqa_diamond=66.7,
```
anyacherniss, line 787: These scores are for the thinking model, however the name of the model we're using does not specify it is thinking (however we do specify thinking vs non models in the app currently, ex. posted below). 

CleanShot 2025-06-18 at 11 18 55@2x

guillaq: @anyacherniss there was a difference in price before hence the 2 different model IDs. It's no longer the case so the model is thinking by default.

We are currently changing the way we handle reasoning to make it a "separate" parameter, follow up in https://linear.app/workflowai/issue/WOR-4986/follow-up-ux-for-reasoning-effort-and-budget

api/core/domain/models/model_datas_mapping.py
guillaq: ✅

PR #499, opened 2025-06-18
Feature/structured model supports

pierrevalade:
sorry my internal testing are not working yet (don't have mongodb setup) so I'll make this PR as a draft.
@guillaq thank you! 

the main goals of this API are:
- use a Pydantic object to represent the `support` object returned in `/v1/models`
- remove fields that are not meant to be exposed via `/v1/models`, for example

```
"support_system_messages":true,"structured_output":true,"support_input_schema":true
```

@guillaq if Jacek needs access to these fields, we need another API endpoint I think, or at least a query parameter on `/v1/models` (probably not exposed publicly).


pierrevalade:
@guillaq should we keep this PR or close? thanks

pierrevalade:
@guillaq

PR #501, opened 2025-06-18
MCP Review: June 18

pierrevalade:
notes collected while exploring our current MCP server.

PR #502, opened 2025-06-18
Implement MVP of search runs tool

yannbu:
Still WIP but works:

Screenshot 2025-06-18 at 23 08 33

### DIFF
--- a/api/api/routers/mcp/_mcp_models.py
+++ b/api/api/routers/mcp/_mcp_models.py
@@ -485,3 +485,20 @@ class StandardModelResponse(BaseModel):
             )
 
     data: list[ModelItem]
+
+
+class RunSearchResult(BaseModel):
+    """Model for run search results"""
+
+    id: str
+    agent_id: str
+    agent_schema_id: int
+    status: str
+    agent_input: str | None
+    agent_output: str | None
+    duration_seconds: float | None
+    cost_usd: float | None
+    created_at: str | None
+    user_review: str | None
+    ai_review: str | None
+    error: dict[str, Any] | None
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -15,6 +15,7 @@ from api.routers.mcp._mcp_models import (
     MajorVersion,
     MCPToolReturn,
     PaginatedMCPToolReturn,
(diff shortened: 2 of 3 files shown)


api/api/routers/mcp/_mcp_models.py
```diff
@@ -485,3 +485,20 @@ def from_model_data(cls, id: str, model: FinalModelData):
             )
 
     data: list[ModelItem]
+
+
+class RunSearchResult(BaseModel):
```
pierrevalade, line 490: see also:
> https://github.com/WorkflowAI/WorkflowAI/pull/501/files#diff-e7e24a283807cea36415003b404fd407dc72c8a00edc3b102f0337d6b93af238R264

I do think that we need two different models: one for a `RunSearchResult` and other model for a Run that has more information. 👍

api/api/routers/mcp/_mcp_models.py
```diff
@@ -485,3 +485,20 @@ def from_model_data(cls, id: str, model: FinalModelData):
...
+    """Model for run search results"""
+
+    id: str
+    agent_id: str
+    agent_schema_id: int
```
pierrevalade, line 495: missing a few important fields that are listed [here](https://github.com/WorkflowAI/WorkflowAI/pull/501/files#diff-e7e24a283807cea36415003b404fd407dc72c8a00edc3b102f0337d6b93af238R264)
for example `agent_version_id`, ...

maybe we should have a `RunBaseModel` with all the information that needs to be returned every time for all runs, no matter the context.

api/api/routers/mcp/mcp_server.py
```diff
@@ -317,8 +330,8 @@ async def get_agent_versions(
     return await service.list_agent_versions(task_tuple, page=page)
 
 
-# @_mcp.tool() WIP
```
yannbu, line 320: I had an early version that was not working at all and was commented.

api/api/routers/mcp/_mcp_models.py
yannbu: @pierrevalade I've added the agent version ID. I think @guillaq  is reworking the run models, so I will merge as this in my branch and potentially adjust later.

PR #509, opened 2025-06-19
Implement elegant sorting solution and tests

pierrevalade:
The MCP server's sorting mechanism was refactored to use a two-field approach: `sort_by` and `order`. This replaces the previous single, monolithic sort parameter, offering a more flexible and standard API design.

*   In `api/api/routers/mcp/_mcp_models.py`, new type aliases `AgentSortField`, `ModelSortField`, and `SortOrder` were introduced to define valid sort fields and directions (`"asc"`, `"desc"`).
*   The `sort_agents` function in `api/api/routers/mcp/_utils/agent_sorting.py` and `sort_models` in `api/api/routers/mcp/_utils/model_sorting.py` were updated to accept both `sort_by` and `order` parameters. Sorting logic now dynamically applies ascending or descending order based on the `order` field, while maintaining stable secondary sorting by ID.
*   The `list_available_models` and `list_agents` tool parameters in `api/api/routers/mcp/mcp_server.py` were updated to reflect the new `sort_by` and `order` fields, including default values.
*   Corresponding method signatures and calls in `api/api/routers/mcp/_mcp_service.py` were adjusted.
*   Unit tests in `api/api/routers/mcp/_utils/agent_sorting_test.py` and `api/api/routers/mcp/_utils/model_sorting_test.py` were comprehensively updated to validate the new two-field sorting behavior across all fields and orders.

This change provides a clearer, more extensible, and type-safe sorting API.

### DIFF
--- a/api/api/routers/mcp/_utils/agent_sorting_test.py
+++ b/api/api/routers/mcp/_utils/agent_sorting_test.py
@@ -99,7 +99,7 @@ class TestSortAgents:
         sorted_agents = sort_agents(agents, "last_active_at", "desc")
 
         # Agents with dates come first (sorted by date desc), then agents without dates (sorted by agent_id asc since they have same date)
-        assert [a.agent_id for a in sorted_agents] == ["agent2", "agent4", "agent1", "agent3"]
+        assert [a.agent_id for a in sorted_agents] == ["agent2", "agent4", "agent3", "agent1"]
 
     def test_sort_by_last_active_at_stable_ordering(self):
         """Test that sorting is stable for agents with same/no dates."""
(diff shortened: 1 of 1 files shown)


api/api/routers/mcp/_utils/agent_sorting.py
```diff
@@ -1,44 +1,63 @@
...
+
+            # For descending order, negate the timestamp while keeping agent_id ascending
+            if order == "desc":
+                return (-max_timestamp, agent.agent_id)
+            # For ascending order, use timestamp as-is
+            return (max_timestamp, agent.agent_id)
+
+        agents.sort(key=get_sort_key)
+    elif sort_by == "total_cost_usd":
+        if order == "desc":
```
yannbu, line 49: should probably use `reverse=...` like for model sorting, instead of a IF ?

yannbu:
thanks @pierrevalade I've tested the code and simplified a bit our merge to my branch.

PR #517, opened 2025-06-19
Create june18-prompt-cursor.md

pierrevalade:
@guillaq @yannbu a work-in-progress of my attempts of prompting CursorAI.

### DIFF
--- a/.discussions/june18-prompt-cursor.md
+++ b/.discussions/june18-prompt-cursor.md
@@ -1,0 +1,39 @@
+<PROMPT>
+...
+</PROMPT>
+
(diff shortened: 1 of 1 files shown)


guillaq:
I think the commands already are the .cursor/rules for python https://github.com/WorkflowAI/WorkflowAI/blob/main/.cursor/rules/python.mdc ?

pierrevalade:
Yeah. I think we need to find a way that the rule be only applied if the request is coming from the background agent. Let me check..... Does not look like there is a way out of the box. Maybe we could use the Does not look like there is a way out of the box. Maybe we could use [.environment](https://docs.cursor.com/background-agent) file, but I think we should check later!

PR #520, opened 2025-06-19
AI Engineer: Prompt, TODOs, Resources

pierrevalade:
Proposing a new architecture for our AI engineer.

### DIFF
--- a/docsv2/content/docs/use-cases/new_agent.mdx
+++ b/docsv2/content/docs/use-cases/new_agent.mdx
@@ -74,8 +74,8 @@ WorkflowAI exposes an API endpoint that is 100% compatible with the OpenAI API `
 
 So, in order to build a new agent, you can use the OpenAI SDK, for example:
 
-<!-- TODO: give code samples for other languages than Python. -->
-<!-- TODO: update URL to run.workflowai.com -->
+{/* TODO: give code samples for other languages than Python. */}
+{/* TODO: update URL to run.workflowai.com */}
 
 ### Example: One-off Processing Agent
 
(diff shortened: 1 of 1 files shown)


.discussions/june19-ai-engineer-architecture.md
```diff
@@ -0,0 +1,89 @@
+# AI Engineer Agent architecture
+
+- [ ] The agent should have access to the following tools:
+- [ ] `search_documentation(query)` to search the documentation of the WorkflowAI platform (1). Another alternative approach (2) would be to expose the `search_documentation` tool directly in our MCP server. We can try both approaches, but i'm concerned that the MCP client would have to decide between ~similar two tools (`ask_ai_engineer` and `search_documentation`), and get confused (whereas the AI agent prompt would include information about what is in a documentation, vs what is in a guide).
+
+- [ ] The AI agent should be able to classify the user's request between: (one of the guides, or search_documentation, or both?) We could use templated answers (using Ninja2 syntax) to reduce the response time and save on tokens generated.
```
guillaq, line 6: By templated answer, you mean we would use structured output to get the model to return the variables for a template ?

guillaq (commented):
I like the idea of using templates.

The guide for a new agent is quite extensive. I wonder if we should return the overall steps but detail only the first step. Cursor could always call the ai engineer agent with details about the second step.

Idea would be to reduce possible token count and possibility of confusion of cursor.

.discussions/june19-ai-engineer-architecture.md
pierrevalade: > By templated answer, you mean we would use structured output to get the model to return the variables for a template ?

see the prompt below. I meant that the LLM was use variables in a Ninja2 template to avoid burning tokens by "pasting" a long section of our guides.

yannbu:
I think the strategic question we need to ask ourselves regarding this whole MCP server project. Is whether we just want to expose to Cursor all the right tools to do the job itself (in this case the "AI engineer" is more like a `search_docs` agent), or do we want also to have a super powerful AI engineer that does things on his side and comes back to Cursor.

pierrevalade:
> I think the strategic question we need to ask ourselves regarding this whole MCP server project. Is whether we just want to expose to Cursor all the right tools to do the job itself (in this case the "ask AI engineer" is more like a `search_docs` agent), or do we want also to have a super powerful AI engineer that does things (calls tools) on his side before answering to Cursor.

good question, right now I want us to explore the **second** direction

well, actually, after today's discussion with @guillaq and @yannbu i'm honestly not sure at all between the two paths. i think we should stay flexible and explore.

guillaq (approved):

PR #525, opened 2025-06-20
Add metadata to RunV1 object

guillaq:
The `RunV1` API model was updated to include a `metadata` field.

*   A `metadata: dict[str, Any] | None` field was added to the `_BaseRunV1` class in `api/api/routers/runs_v1.py`, which is inherited by `RunV1` and `RunItemV1`.
*   The `from_domain` and `from_domain_task_run` conversion methods in `api/api/routers/runs_v1.py` were modified to populate this `metadata` field from the underlying `AgentRun` domain model, which already contained this information.
*   This change ensures that user-defined key-value pairs stored with an `AgentRun` are now exposed via the `/v1/runs` API endpoints.
*   Comprehensive tests were added to `api/api/routers/runs_v1_test.py`, including `test_latest_run_includes_metadata` and `test_latest_run_with_null_metadata`, to verify correct metadata inclusion and handling of `None` values (which are excluded from the response due to `response_model_exclude_none=True`).

### DIFF
--- a/api/api/routers/runs_v1.py
+++ b/api/api/routers/runs_v1.py
@@ -97,6 +97,11 @@ class _BaseRunV1(BaseModel):
         description="A signed token that can be used to post feedback from a client side application",
     )
 
+    metadata: dict[str, Any] | None = Field(
+        default=None,
+        description="A user set metadata key / value. Keys are not searchable.",
+    )
+
 
 class RunItemV1(_BaseRunV1):
     task_input_preview: str = Field(description="A preview of the input data")
@@ -205,6 +210,7 @@ class RunV1(_BaseRunV1):
             ),
             feedback_token=feedback_token,
             conversation_id=run.conversation_id,
+            metadata=run.metadata,
         )
 
 
@@ -231,7 +237,7 @@ async def get_run(
     runs_service: RunsServiceDep,
     feedback_token_generator: RunFeedbackGeneratorDep,
 ) -> RunV1:
-    run = await runs_service.run_by_id(task_tuple, run_id)
+    run = await runs_service.run_by_id(task_tuple, run_id, exclude={"llm_completions"})
     return RunV1.from_domain_task_run(run, feedback_token_generator(run.id))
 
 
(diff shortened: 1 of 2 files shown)


pierrevalade:
@guillaq when ready, I'll review the PR as a API change, thank you.

api/api/routers/runs_v1.py
```diff
@@ -181,6 +181,11 @@ def from_domain(cls, error: ErrorResponse.Error):
 
     conversation_id: str | None
 
+    metadata: dict[str, Any] | None = Field(
```
guillaq, line 184: @pierrevalade only API change. 
Added metadata field to RunV1 object which is returned by `/{agent_id}/runs/latest` and the run by ID endpoint

api/api/routers/runs_v1.py
```diff
@@ -181,6 +181,11 @@ def from_domain(cls, error: ErrorResponse.Error):
...
+        description="Combination of user defined metadata passed to the completion request, and system metadata "
```
pierrevalade, line 185: Please clarify what system metadata are added. I also don't like the term "system" metadata (but I don't know what metadata are added, either).

"system" has a particular meaning in LLM

api/api/routers/runs_v1.py
```diff
@@ -181,6 +181,11 @@ def from_domain(cls, error: ErrorResponse.Error):
...
+        "added by WorkflowAI",
```
pierrevalade, line 186:
```suggestion
        "automatically added by WorkflowAI",
```

pierrevalade (changes requested):

api/api/routers/runs_v1.py
pierrevalade: @guillaq you can see the work I started here for exposing hosted tools via the API, and MCP.
> https://github.com/WorkflowAI/WorkflowAI/pull/530

I don't think we need a tool in the MCP server to list all of WorkflowAI's added metadata. I do think that we should probably have a reference in the code, maybe a list of Pydantic objects.
And the list of added metadata wouldn't be a bad idea to include in the API and then have the documentation fetch dynamically the list of added metadata from the API endpoint.

pierrevalade: OK, I'll open a Linear to clarify what `workflowai.` metadata are added automatically.

pierrevalade (approved):

pierrevalade (approved):

PR #527, opened 2025-06-20
Add cursor env

guillaq:
ref https://linear.app/workflowai/issue/WOR-5009/cursorenvironmentjson


pierrevalade (approved):

guillaq:
Bugbot is mistaken but I can't blame it. We have two api directories

PR #533, opened 2025-06-22
feat: add metadata_fields to MCP get_agent for search filtering discovery

pierrevalade:
## Problem

The MCP search tool allows filtering by metadata fields (e.g., `metadata.user_email`, `metadata.environment`) but provides no way to discover what metadata fields are available. Users cannot determine which metadata fields exist because the metadata schema is not exposed in the MCP tool response.

## Solution

Added `metadata_fields` array to the `AgentResponseDetailed` model returned by the `get_agent` MCP tool. This provides a discovery mechanism where users can:

1. Call `get_agent` to get detailed agent information
2. Receive a list of available metadata field paths (e.g., `['metadata.user_email', 'metadata.environment', 'metadata.session_id']`)
3. Use these exact field paths in `search_runs` for filtering

## Changes

- **Model**: Added `metadata_fields: list[str]` to `AgentResponseDetailed` with Pydantic field description
- **Service**: Added placeholder with TODO comment for Guillaume to implement the discovery logic  
- **Documentation**: Updated `get_agent` tool docstring to explain the new feature

## Implementation Status

✅ **Ready**: API contract defined, model changes complete
🔄 **TODO**: Guillaume to implement actual metadata field discovery logic by sampling recent runs

## Benefits

- **Discoverability**: Users know what metadata fields exist before searching
- **Better UX**: No more guessing field names or trial-and-error
- **Context-aware**: Metadata fields are specific to each agent
- **Non-intrusive**: Enhances existing `get_agent` tool without breaking changes

## Testing

Once Guillaume implements the discovery logic, users will be able to:
1. Call the MCP `get_agent` tool 
2. See available metadata fields in the response
3. Use those exact field paths in `search_runs` queries

Related issue: MCP Search Tool Lacks Metadata Field Discovery

### DIFF
--- a/docsv2/content/docs/observability/runs.mdx
+++ b/docsv2/content/docs/observability/runs.mdx
@@ -1,5 +1,6 @@
 ---
 title: Runs
+summary: Learn how to track, view, and search AI agent runs, including how to use metadata for debugging and troubleshooting.
 ---
 
 import { Tabs, Tab } from "fumadocs-ui/components/tabs"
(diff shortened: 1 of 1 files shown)


pierrevalade:
@guillaq Great point! You're absolutely right that we should expose all search fields, not just metadata fields.

Instead of creating a separate MCP tool, I think we should expand the current approach in `get_agent` to include **all searchable fields** in a single attribute. This would give users complete discovery of what they can search for without needing multiple tool calls.

**Proposed solution:**
- Rename `metadata_fields` → `searchable_fields` 
- Include all search parameters that the `search_runs` tool accepts in one flat array

Example response:
```json
{
  "agent_id": "customer-support-agent",
  "searchable_fields": [
    "status",
    "time", 
    "price",
    "latency",
    "model",
    "schema",
    "version",
    "temperature",
    "source",
    "review",
    "metadata.user_id",
    "metadata.user_email",
    "metadata.environment",
    "input.message",
    "input.customer_id",
    "output.response",
    "output.classification"
  ],
  // ... other agent details
}
```

This way, a single `get_agent` call gives users the complete "search vocabulary" for that agent, making the search experience much more discoverable. The field names are self-explanatory (metadata.*, input.*, output.*, and run properties).

What do you think? Should I update the PR to expand the scope to all searchable fields?

guillaq:
> run_search_fields
> This makes it crystal clear these are the exact field paths you can use in search_runs queries.

Sounds good to me !

PR #540, opened 2025-06-22
Fix agent ID inconsistency: preserve underscores in slugify function

pierrevalade:
## Summary

Fixes agent ID inconsistency where underscores were being converted to hyphens, causing mismatches between user code and WorkflowAI URLs.

## Problem
- User code: `metadata={"agent_id": "text_summarizer_agent"}`  
- WorkflowAI URL: `https://preview.workflowai.com/workflowai/agents/text-summarizer-agent/1`
- **Inconsistent**: underscores vs hyphens

## Solution
Updated the `slugify` function to intelligently preserve existing separators:
- ✅ **Preserves underscores**: `text_summarizer_agent` → `text_summarizer_agent`
- ✅ **Preserves hyphens**: `kebab-case-text` → `kebab-case-text`  
- ✅ **Handles mixed**: `mixed-CASE_string` → `mixed-case_string`
- ✅ **Backward compatibility**: `Hello World` → `hello-world`

## Changes
- Modified `core/utils/strings.py::slugify()` function
- Updated corresponding tests in `core/utils/strings_test.py`
- All tests passing ✅
- Ruff checks passing ✅

## Testing
- Added test cases for the Linear issue scenarios
- Updated existing tests to reflect new behavior for underscore-containing inputs
- Verified backward compatibility with existing use cases

Closes [WOR-5049](https://linear.app/workflowai/issue/WOR-5049/agent-id-inconsistency-code-uses-underscores-while-workflowai-api-uses)

---
*Note: This code was generated with AI assistance.*


api/core/utils/strings.py
```diff
@@ -56,8 +56,26 @@ def normalize(s: str, case_sensitive: bool = False, remove_punctuation: bool = T
 
 def slugify(s: str) -> str:
     n = normalize(s, case_sensitive=True, remove_punctuation=False)
+
+    # Preserve existing separators by splitting on them, processing parts, then rejoining
```
pierrevalade, line 60: **FYI: vibe coded.. I did not check at this code change, just focused on the tests passing.**

guillaq:
Would it not be better to always convert the agent_id to a slug before processing  ? 

For example, `have https://preview.workflowai.com/workflowai/agents/text-summarizer-agent/1` and 
`https://preview.workflowai.com/workflowai/agents/text_summarizer_agent/1` be equivalent ?

Is there any benefit in distinguishing `test_summarizer_agent` and `test-summarizer-agent` ?

guillaq (changes requested):
Avoiding to "fix" valid fields sounds good to me for the reasons listed above.

We should then not modify the `slugify` function since `slugify` has a clear meaning and may be used elsewhere. The `fail fast` with the `is_url_safe` means that people with "existing code" with spaces in their agent name would trigger errors. Do we want to let cursor fix the issue ? I think that's ok since the error is clear enough ?

pierrevalade:
Great points about keeping `slugify()` unchanged! I agree that maintaining its clear semantic meaning is important, especially since it may be used elsewhere.

The fail-fast validation approach with helpful error messages sounds like the right solution. To make this even better for developers (and AI assistants like Cursor), I think the error message should include the correctly formatted agent ID.

**Example error message:**
```
Invalid agent_id: "Text Summarizer Agent" contains invalid characters. 
Agent IDs must be URL-safe (lowercase letters, numbers, hyphens, and underscores only).

Suggested fix: Use "text-summarizer-agent" instead.
```

**Benefits of this approach:**
1. ✅ **Clear actionable feedback** - developers know exactly what to change
2. ✅ **AI-friendly** - tools like Cursor can easily parse and auto-fix these errors  
3. ✅ **Preserves slugify semantics** - no unintended side effects elsewhere
4. ✅ **Explicit over implicit** - no surprises about URL formatting

The error can call `slugify(agent_id)` to generate the suggested replacement, making it trivial for developers to copy-paste the correct format.

**Implementation Checklist:**
- [ ] **Accept underscores as valid**: Ensure `is_url_safe()` accepts agent IDs with underscores (e.g., `text_summarizer_agent` should be valid)
- [ ] **Update validation logic**: Modify validation to allow underscores while still rejecting spaces and other invalid characters
- [ ] **Error message includes slugified suggestion**: Call `slugify(agent_id)` to show the correct format for invalid IDs
- [ ] **Review unit tests**: Ensure test cases properly validate that underscores are accepted and error messages are helpful

@Guillaume Could you review the current unit test error message for invalid agent IDs? I'd like to make sure we're providing the most helpful error format and that we're correctly accepting underscores as valid.

This way we get the best of both worlds: strict validation that prevents confusion, plus helpful guidance that makes fixing the issue effortless.

pierrevalade:
@guillaq looks like Cursor like to generate `agent_id` like `email_urgency_classifier` (with hyphens), so could you please have a look at this issue tomorrow Wed or Thursday at the latest? thank you!

PR #545, opened 2025-06-23
Uniformize parameter formats in mcp_server.py

yannbu:
The parameter format in `api/api/routers/mcp/mcp_server.py` has been uniformized to exclusively use Pydantic `Field` directly instead of Annotated.

Annotated is too verbose and not required IMO.

### DIFF
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -1,4 +1,4 @@
-from typing import Annotated, Any, Literal
+from typing import Any, Literal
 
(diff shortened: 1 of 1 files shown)


pierrevalade:
OK, @guillaq will decide when to merge, there are already many pull requests open from over the weekend that haven't been merged.

guillaq:
Not sure which one is better between annotations or default values for Fields. I believe FastMCP recommands annotations without justification + default values is somewhat less verbose and better indented.

Confirm that default values work fine. This PR should be rebased on main though

PR #558, opened 2025-06-23
Documentation: Playground modules consolidation

anyacherniss:
The intended scope of this PR is to just clean up the pages, as requested on Slack (see screenshot of message for context). 

There is still work to do (denoted by "TODO"s) to actually polish each section, but for now I am considering that out of scope and will handle in a separate PR. 

CleanShot 2025-06-23 at 17 42 47@2x


anyacherniss:
@pierrevalade I'm going to close this PR and push a new PR with the playground module consolidation (among some other changes). I think I was overthinking when I made this PR, going for simpler organization of features.

Since this PR hasn't been reviewed (and has a merge conflict??) I don't think there is anything particular of value that justifies keeping it open

PR #565, opened 2025-06-24
Revert to local docs mode

yannbu:


api/api/routers/mcp/mcp_server.py
```diff
@@ -688,97 +688,97 @@ async def search_documentation(
...
-     - 'use-cases/image-input' - Use case guide for using images as input for AI agents. Covers processing and analyzing images to extract information, answer questions, and perform vision-based tasks.
-     - 'use-cases/pdf-input' - Use case guide for processing and extracting information from PDF documents. Covers techniques for using PDF files as input for summarization, data extraction, and question-answering tasks.
-     - 'use-cases/image-generation' - Use case guide for generating and editing images. Covers best practices for prompting, customizing image outputs, and using the WorkflowAI SDK for image generation tasks.
-     - 'use-cases/mcp' - A collection of use-case scenarios for AI agents that interact with the WorkflowAI platform. Outlines the required capabilities for agents to perform tasks like optimization, debugging, and deployment.
+     - 'use-cases/chatbot.mdx' - Use case guide for building a conversational AI chatbot. Covers basic setup, conversation management, streaming, error handling, and deployment.
```
guillaq, line 702: Why do we need the `.mdx` ? That seems confusing. Can we auto add if needed in local mode instead ?

yannbu: done https://github.com/WorkflowAI/WorkflowAI/pull/565/commits/80c58c9400fa03dedc49e466cc40652a46ed3f5b

PR #568, opened 2025-06-24
[tentative] add link to stripe fulfillment policy on landing page

anyacherniss:
Note: unsure if this is needed, but ChatGPT strongly recommends linking the fulfillment policy on the landing page. This isn't any super pretty UI, just asked Cursor to add a basic footer with the link.

Ref: https://linear.app/workflowai/issue/WOR-5058/stripe-fulfillment-policy

https://github.com/user-attachments/assets/f15abd67-6bce-48b4-b38b-a1dacb4b952f

### DIFF
--- a/client/src/app/landing/LandingPage.tsx
+++ b/client/src/app/landing/LandingPage.tsx
@@ -155,6 +155,18 @@ export function LandingPage() {
           className='sm:mt-20 mt-14 sm:mb-20 mb-14'
           routeForSignUp={routeForSignUp}
         />
+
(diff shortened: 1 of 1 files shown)


anyacherniss:
Note that _if_ it's determined that we need this change (I've asked Nina to confirm if simply having the policy in the documentation is sufficient), this would need to be a hotfix. Do I need to rename the branch in that case @guillaq? (Sorry for being ignorant on this topic. I feel like I should know this by now)

pierrevalade:
no we're not putting that link here on our homepage.

PR #581, opened 2025-06-25
retry_run tool

yannbu:
Related to [WOR-4993: Investigate repeated quote in GenerateSoapnoteFromTranscript output](https://linear.app/workflowai/issue/WOR-4993/investigate-repeated-quote-in-generatesoapnotefromtranscript-output)

While investigating how to fix the issue above with MCP, I found that Cursor struggles to know when to use `create_completion` with a `orginal_run_id`, e.g

Screenshot 2025-06-25 at 16 41 46

Cursor is stuck regenrating the transcript it got from a previous `fetch_run_details`.

With my update, the behaviour changes. Cursor understands it must call the `retry_run` tool and still struggles to understand that it must pass an updated response format —since this is what it had just fixed in the user code. But thanks to the explicit error, Cursor corrects its tool call. 

Screenshot 2025-06-25 at 17 01 48

(sorry, I should have copied the full markdown of the Cursor discussion for clarity. I'll do that next time.)

I'm not 100% sure we need a separate `retry_run` tool, but this is an option we have to make the usage more obvious.

### DIFF
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -32,6 +32,7 @@ from api.routers.mcp._mcp_serializer import tool_serializer
 from api.routers.openai_proxy._openai_proxy_models import (
     OpenAIProxyChatCompletionRequest,
     OpenAIProxyChatCompletionResponse,
+    OpenAIProxyMessage,
 )
 from api.services.documentation_service import DocumentationService
 from api.services.tools_service import ToolsService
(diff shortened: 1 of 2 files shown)


api/api/routers/mcp/mcp_server.py
```diff
@@ -710,6 +711,71 @@ async def create_completion(
...
+async def retry_run(
+    # TODO: we should not need the agent id here
+    agent_id: str = Field(
+        description="The id of the user's agent. Example: 'agent_id': 'email-filtering-agent' in metadata, or 'email-filtering-agent' in 'model=email-filtering-agent/gpt-4o-latest'.",
+    ),
+    original_run_id: str = Field(
+        description="The id of the run to retry.",
+    ),
+    model: str = Field(
+        description="The model to use for the retry. Even if the same model is used/",
```
pierrevalade, line 724: why do we need "Even if the same model is used"?

api/api/routers/mcp/mcp_server.py
```diff
@@ -710,6 +711,71 @@ async def create_completion(
...
+    ),
+    temperature: float | None = Field(
+        description="The temperature to use for the retry. In case the temperature is not the same as the original run.",
+        default=None,
+    ),
+) -> MCPToolReturn[OpenAIProxyChatCompletionResponse]:
+    """Retry a run. Prefer to use this tool over creating a new run, especially when the input and messages are long.
+
+    <when_to_use>
+    When the user wants to retry a run for example after fixing a bug in their agent (in this case the 'original_run_id' is likely a run that failed and that they want to retry after updating their agent).
```
pierrevalade, line 742: see this commit: https://github.com/WorkflowAI/WorkflowAI/pull/574/commits/5d61caf30fe90a4effc948ad936a6d8cd7441bdc on how to edit the description to remove "user" references.

pierrevalade:
continued in https://github.com/WorkflowAI/WorkflowAI/pull/639

PR #586, opened 2025-06-25
Updated documentation links for proxy meta agent

anyacherniss:
Closes: https://linear.app/workflowai/issue/WOR-5039/update-proxy-meta-agent-documentation-links

Searched the files for other instances of older URLs, but looks like these are the only ones ✅

### DIFF
--- a/api/api/services/internal_tasks/meta_agent_service.py
+++ b/api/api/services/internal_tasks/meta_agent_service.py
@@ -1858,7 +1858,7 @@ Please double check:
                     yield [
                         MetaAgentChatMessage(
                             role="ASSISTANT",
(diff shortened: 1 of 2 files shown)


anyacherniss:
Sorry for the false alarm re: code review, please ignore for now! Still seeing the old documentation link sent when testing the proxy agent chat - need to figure out if it's just some delay, or if I'm missing something.

pierrevalade (commented):
@anyacherniss should we use `docs.*` domain knowing that the new docs will be soon published to this domain?

The links won't work until we move to the new documentation. But at least we don't have to update the links later.

pierrevalade (approved):

guillaq (approved):

PR #590, opened 2025-06-25
Add additional playground documentation

anyacherniss:
As discovered when investigating https://linear.app/workflowai/issue/WOR-5089/playground-chat-not-aware-of-playground-deployments, our current documentation doesn't mention deployments being available on the playground.


anyacherniss:
Closing in favor of a new PR that is a more practical consolidation of the documentation updates I need to make

PR #603, opened 2025-06-26
Playground documentation updates

anyacherniss:
Goal of this PR:
- Consolidate playground documentation modules into one, clean "features" page
- Add details about how to deploy from the playground
- Add steps on how schemas can be updated (ie. not on the playground). Guessing at this point that I will need to put additional instructions in the proxy meta agent directly.


pierrevalade:
thank you @anyacherniss, I need to think a bit more about how to introduce the playground in our documentation, I still haven't completely wrapped my head around all the implications of MCP and the collaboration between MCP and the playground, so I need to think more about MCP in a couple days. Thank you.

PR #611, opened 2025-06-27
Enhanced model and output structure for the MPC observer.

yannbu:
@guillaq , I noticed some generation failures with Gemini 2.0 flash, so I reworked the output schema and the model.

### DIFF
--- a/api/api/routers/mcp/_mcp_observer_agent.py
+++ b/api/api/routers/mcp/_mcp_observer_agent.py
@@ -9,16 +9,20 @@ from pydantic import BaseModel, Field
 
 class MCPToolCallObserverOutput(BaseModel):
     success: bool = Field(description="Whether the feedback was processed successfully")
-    error_origin: Literal["mcp_client", "our_mcp_server"] | None = Field(
-        description="The origin of the error, if any. 'mcp_client' are typically wrong arguments passed etc, and are less critical. 'our_mcp_server' are more serious errors, like the server not being able to process the request.",
-        default=None,
-    )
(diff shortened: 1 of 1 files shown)


api/api/routers/mcp/_mcp_observer_agent.py
```diff
@@ -61,7 +65,7 @@ async def mcp_tool_call_observer_agent(
     # TODO: add some mcp-session_id to be able to fetch the full MCP conversation too
 
     response = await client.beta.chat.completions.parse(
-        model="gemini-2.0-flash-latest",
+        model="gpt-4.1-nano-latest",  # Cheap, native structured generation.
```
pierrevalade, line 68: I don't think we are really optimizing on cost, here, but instead collecting quality data points, especially as the MCP feature will be in the hands of users next week. why not pick a GREAT model?

yannbu: OK, I've switched to Mini 20$ / 1k runs. I could go GPT 4.1 full but that is 160$ per 1k runs. WDYT ?

yannbu: I'm limiting to GPT 4.1 models because I want large context window (at least 1M) and native structured generation.

guillaq (approved):

api/api/routers/mcp/_mcp_observer_agent.py
yannbu: o3 has only 200k context size. 
Switched to 4.1 full

PR #613, opened 2025-06-27
update MCP middleware session management

yannbu:
@guillaq I believe this would solve the session management problems I encountered in dev. 

And I've added logs so we can know more, if the problem is not fixed.


yannbu:
> Not sure I get a change but good to try.
> 
> Are you manually creating a session id ? Are you sure Cursor picks up on that header ?

@guillaq behaviour it looks a bit transient. When I opened this MR, the MCP session ID was always given back when I created it, but today I've tried again today and it's not. So I'd say we merge and we observe the logs.

PR #615, opened 2025-06-27
Add pricing to tool objects and simplify tool definitions

pierrevalade:
## Summary

This PR adds pricing information to WorkflowAI tools and simplifies the tool definition architecture by removing unnecessary abstractions.

### 1. New Centralized Tool Definitions
- Created `api/core/tools/tool_definitions.py` with simple Pydantic-based tool definitions
- Each tool now has a `name`, `description`, and `price` field
- All tool metadata is centralized in one location for easy maintenance

### 2. Added Price Field
- Added `price: str | None` field to the `Tool` domain model
- Using string type for pricing flexibility (to be interpreted by agents)
- Added price field to `HostedToolItem` in both MCP and REST API responses

### 3. Simplified Architecture
- Removed unnecessary `ToolsService` abstraction for hosted tools
- Both MCP server and tools v1 router now use `AVAILABLE_TOOLS` directly
- Clear separation between hosted tools (in `tool_definitions.py`) and custom tools (still managed by `ToolsService`)

### 4. Updated Documentation
- Updated code comments to reflect that tool descriptions are now centralized
- Updated API documentation to mention pricing information
- Added TODO in docs to include pricing in the tools table

## TODO
- Replace `` placeholders with actual pricing values in `tool_definitions.py`

## Testing
- ✅ All ruff checks pass
- ✅ All pyright type checks pass
- ✅ No breaking changes to existing APIs

### DIFF
--- a/docsv2/content/docs/agents/tools.mdx
+++ b/docsv2/content/docs/agents/tools.mdx
@@ -41,13 +41,13 @@ WorkflowAI currently supports tools for:
 
 To add a hosted tool to your agent, find the tool name (`@tool`) and add it to your agent prompt.
 
-For example, to use the `@google-search` tool, you can add the following to your agent prompt:
+For example, to use the `@search-google` tool, you can add the following to your agent prompt:
 
 ```python
 messages = [
 {
   "role": "system",
-  "content": "Use @google-search to find the weather in {{location}}."
+  "content": "Use @search-google to find the weather in {{location}}."
 },
 ]
 ```
@@ -67,7 +67,7 @@ Tools can be added in the Playground by either:
 ### Web search
 
 WorkflowAI supports two web search tools:
-- `@google-search` makes a web search using Google
+- `@search-google` makes a web search using Google
 - `@perplexity-sonar-pro` makes a web search using Perplexity's Sonar Pro model
 
 <Tabs items={["Google Search", "Perplexity"]}>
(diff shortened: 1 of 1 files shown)


api/core/tools/tool_definitions.py
```diff
@@ -0,0 +1,83 @@
...
+
+class GoogleSearchTool(BaseTool):
+    """Google search tool definition."""
+
+    name: str = "@search-google"
+    description: str = (
+        "Performs a Google web search using Serper.dev API and returns search results "
+        "including links, snippets, and related information in JSON format."
+    )
+    price: str = "$TODO per call (search)"
```
pierrevalade, line 27: I've opened a Linear issue to get all the prices information:
> https://linear.app/workflowai/issue/WOR-5125/define-pricing-for-workflowai-hosted-tools

pierrevalade:
@guillaq ready for a first pass. I don't like that we have tools defined into two different places still, but I prefer the suggested Pydantic-based approach. what do you think?

pierrevalade:
bugbot run

guillaq (commented):
Right now we have the `build_tool` function that exports a BaseModel `Tool` object manually and in different spots.

The issue with the proposed implementation is that we are using pydantic models as consts (i-e only using the default field values) which is somewhat overkill.

Because the tools we have essentially have a different pricing structure, we won't be able to leverage a common "ToolPricing" definition I think.

However, I really like the new Class based approach. Basically, instead of being a single function `browser_text`, etc. A tool is now a full fledged class with:
- a `Tool` definition object that is explicitly built (as opposed to being dynamically generated)
- a `__call__` function that executes the tool
- a clear price computation that uses explicit consts like the one you set

pierrevalade:
ok, thanks for the review, you're OK to take next steps @guillaq? I agree a simpler Class approach would work actually.

PR #618, opened 2025-06-27
Add max and min budgets

guillaq:


pierrevalade:
@guillaq should `min_reasoning_budget` and `max_reasoning_budget` be added to `ConciseModelResponse` as well, to cover the MCP `list_models` response?

guillaq:
@pierrevalade done !

guillaq:
I don't think we should push the MCP to set a specific reasoning budget. the `disabled` `low` `medium` `high` seem enough to me ?

pierrevalade:
why not? the documentation (available in the `search_documentation`) already mentions the thinking budget in tokens, so the MCP client needs a way to know what are the acceptable values.

guillaq:
Sounds good!

PR #620, opened 2025-06-27
Review search_runs operators: Remove 'is empty' and 'is not empty'

pierrevalade:
## Context

The `search_runs` tool documentation currently includes `"is empty"` and `"is not empty"` operators for both String and Number fields. These operators were added by @YBubu in commit d05e6e15c as part of the MVP search runs tool implementation.

## Issue

I'm not sure these operators are actually supported by the search backend. 

## Changes

- Removed `"is empty" - field has no value` from String fields
- Removed `"is not empty" - field has a value` from String fields  
- Removed `"is empty" - field has no value` from Number fields
- Removed `"is not empty" - field has a value` from Number fields

## Request for Review

**Please review the `operators_by_type` section and confirm:**

1. Are `"is empty"` and `"is not empty"` actually supported operators?
2. If not, are there any other operators in the documentation that should be removed?
3. Are there any missing operators that should be added?

The current operators listed are:
- **String**: is, is not, contains, does not contain
- **Number**: is, is not, greater than, greater than or equal to, less than, less than or equal to  
- **Date**: is before, is after
- **Boolean**: is, is not

@YBubu - since you implemented the original search functionality, could you please review?

### DIFF
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -337,8 +337,6 @@ async def search_runs(
     - "is not" - not equal to
     - "contains" - string contains
     - "does not contain" - string does not contain
-    - "is empty" - field has no value
-    - "is not empty" - field has a value
 
     **Number fields:**
     - "is" - exact match
@@ -347,8 +345,6 @@ async def search_runs(
     - "greater than or equal to" - value >= X
     - "less than" - value < X
     - "less than or equal to" - value <= X
-    - "is empty" - field has no value
-    - "is not empty" - field has a value
 
     **Date fields:**
     - "is before" - date < X
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -333,14 +333,14 @@ async def search_runs(
     Different field types support different operators:
 
     **String fields:**
-    - "is" - exact match
-    - "is not" - not equal to
+    - "is" - exact match (use value "Empty" to check for empty/null fields)
+    - "is not" - not equal to (use value "Empty" to check for non-empty fields)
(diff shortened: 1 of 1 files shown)


pierrevalade:
## Analysis of Search Operators (from Claude 4 Opus)

After investigating the codebase, I found that the 'is empty' and 'is not empty' operators **are actually supported**, but they work differently than the original documentation suggested.

### How Empty Checking Actually Works

1. **The operators exist internally**:  and  are defined in the codebase
2. **But they're accessed via a special value**: Instead of being separate operators exposed to users:
   - Use `"is"` operator with value `"Empty"` → internally converts to `IS_EMPTY`
   - Use `"is not"` operator with value `"Empty"` → internally converts to `IS_NOT_EMPTY`

### Implementation Details

- **Code location**: `api/api/services/runs_search.py` lines 347-352
- **Special keyword**: `_EMPTY_KEYWORD = "Empty"`
- **Storage implementations**: Both MongoDB and ClickHouse have full implementations
- **Tests exist**: `api/core/storage/mongo/utils_test.py` has tests for empty field checking

### Documentation Update

I've updated the documentation to accurately reflect this behavior:
- Removed misleading standalone 'is empty' and 'is not empty' operators
- Added clarification to 'is' and 'is not' operators about using "Empty" value
- Added Example 6 demonstrating proper empty field checking

### All Operators Verified ✓

- **String fields**: is, is not, contains, does not contain
- **Number fields**: is, is not, greater than, greater than or equal to, less than, less than or equal to  
- **Date fields**: is before, is after
- **Boolean fields**: is, is not

All operators match the code implementation in `SearchOperator` class methods.

This preserves existing functionality while providing accurate documentation for users.

guillaq (approved):

PR #624, opened 2025-06-28
Fix context usage to display completion with highest ratio

pierrevalade:
## �� Problem

Context usage display was showing data from the **first completion** instead of the completion with the **highest context usage ratio**.

As reported in the issue, the frontend displayed usage from the first completion (75 tokens) when it should have shown the completion with highest usage (7832 tokens).

## 🔧 Solution

Updated `getContextWindowInformation` function to:
- Calculate usage ratio for each completion: `(prompt_tokens + completion_tokens) / model_context_window_size`
- Select the completion with the **highest ratio** instead of just the first valid one
- Maintain backward compatibility with existing validation logic

## 🧪 Testing

- Added comprehensive test suite with 7 test cases
- Covers edge cases: missing data, multiple completions, equal ratios
- All tests passing ✅
- TypeScript compilation successful ✅  
- ESLint checks passed ✅

## 📸 Before/After

**Before**: Shows first completion (75 tokens) = 1% usage
**After**: Shows highest usage completion (7832 tokens) = 98% usage

This gives users accurate context usage information for optimization.

Fixes **WOR-3301**

### DIFF
--- a/client/src/lib/taskRunUtils.test.ts
+++ b/client/src/lib/taskRunUtils.test.ts
@@ -1,0 +1,188 @@
+import { LLMCompletionTypedMessages } from '@/types/workflowAI';
+import { getContextWindowInformation } from './taskRunUtils';
+
+describe('getContextWindowInformation', () => {
+  it('should return undefined when no completions are provided', () => {
+    expect(getContextWindowInformation(undefined)).toBeUndefined();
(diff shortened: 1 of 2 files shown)


pierrevalade:
bugbot run

anyacherniss (approved):

PR #628, opened 2025-06-29
feat: Add unsupported feature detection to search documentation agent

pierrevalade:
## Summary

This PR enhances the search documentation agent to detect when users are asking about features or capabilities that WorkflowAI doesn't currently support, helping distinguish between missing documentation and unsupported functionality.

## Changes

- **New Model**: Added `UnsupportedFeatureDetection` Pydantic model with:
  - `is_unsupported`: Boolean flag indicating if the query asks about unsupported features
  - `feedback`: Explanatory text about what unsupported feature was requested

- **Enhanced Output**: Updated `SearchDocumentationOutput` to include `unsupported_feature_detected` field

- **Improved Prompting**: Enhanced the system prompt to include task #4 for detecting unsupported features, clearly distinguishing this from missing documentation

## Benefits

- Provides better user experience by clearly identifying when features aren't supported
- Helps distinguish between "we don't have docs for this" vs "we don't support this"
- Enables more targeted feedback for feature requests outside WorkflowAI's scope
- Maintains existing functionality while adding new detection capabilities

## Testing

The changes maintain backward compatibility as the new field is optional with a default value of `None`.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update


yannbu:
@pierrevalade I've experimented the concept this morning and I think the proper way to handle the use case is to have an explicit "unsupported features" documentation pages that recaps the features:
- that are not planned at all
- the features that are "work in progress"
- the future features we are considering (ex: fine tuning, etc...)

Without an explicit doc like that, it's very hard for the docs agent to know whether we are just missing docs or if features are not supported "by design".

We could start with an extensive docs of what we don't support (base on our gut feeling people will ask for and what our users usually ask for). Then as more "missing docs" error will come we can either:
- add the use case to the "unsupported features" documentation
or 
- add a docs

What do you think @pierrevalade ?

pierrevalade:
> @pierrevalade I've experimented the concept this morning with mixed results

Could you please provide more detail about the mixed results? Thank you.

yannbu:
> > @pierrevalade I've experimented the concept this morning with mixed results
> 
> Could you please provide more detail about the mixed results? Thank you.

For example, when asked "Does workflowAI allow to run OpenAI finetuned models ?" the agent decided output:

Screenshot 2025-06-30 at 16 00 03

Notice there is no "unsupported_feature_detected" in the output, but there should be an "unsupported_feature_detected".

So we would need to instruct the agent with either:
- "If some features are undocumented, assume they are unsupported", but in this case why do we need to have separate "unsupported_feature_detected" and "missing_doc_sections_feedback" ?
- "If some features are undocumented, try to guess if they are really unsupported or if only the documentation is missing? " In this case, that would be totally a "random guess from the agent", unless the search documentation agent would have access to our code base. In this case the agent could answer "Does workflowAI allow to run OpenAI finetuned models ?" by seeing that we have a fixed list of model in the code, hence no fine tuning.

pierrevalade:
yes, exactly.

yannbu (approved):

api/api/routers/mcp/_mcp_service.py
```diff
@@ -375,6 +377,25 @@ async def search_documentation(
...
+        if len(relevant_sections) > 0:
+            agent_answer = f"Successfully found relevant documentation sections: {[section.file_path for section in relevant_sections]}."
+        else:
+            agent_answer = "No relevant documentation sections found for your query. Please check the 'foundations' page for general information about WorkflowAI."
+
+        # Add warning about unsupported feature if detected
+        if unsupported_feature_detected:
+            agent_answer = (
+                agent_answer
+                + f" The query is about a feature that WorkflowAI does not currently support: '{unsupported_feature_detected.feedback}'"
```
pierrevalade, line 394: why do we need that part?

api/api/routers/mcp/_mcp_service.py
```diff
@@ -375,6 +377,25 @@ async def search_documentation(
             error="Invalid parameters provided",
         )
 
+    def _get_search_documentation_by_query_message(
+        self,
+        relevant_sections: list[DocumentationSection],
+        unsupported_feature_detected: UnsupportedFeatureDetection | None,
+    ) -> str:
+        if len(relevant_sections) > 0:
```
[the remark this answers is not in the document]
pierrevalade: what was "buggy" exactly?

api/api/routers/mcp/_mcp_service.py
```diff
@@ -375,6 +377,25 @@ async def search_documentation(
...
+            agent_answer = f"Successfully found relevant documentation sections: {[section.file_path for section in relevant_sections]}."
+        else:
+            agent_answer = "No relevant documentation sections found for your query. Please check the 'foundations' page for general information about WorkflowAI."
```
pierrevalade, line 388: I don't understand that answer since the foundations page is always included in the response for a query search?

api/api/routers/mcp/_mcp_service.py
```diff
@@ -396,24 +420,20 @@ async def _search_documentation_by_query(self, query: str) -> MCPToolReturn[Sear
...
-                content_snippet=sections[0].content,
-                source_page="foundations.mdx",
-            ),
-        )
 
-        if len(query_results) == 0:
-            return MCPToolReturn(
-                success=True,
-                message=f"No relevant documentation sections found for query: {query}",
+        if "foundations" not in [section.file_path for section in relevant_sections]:
```
pierrevalade, line 424: did you QA this part properly?

api/core/agents/search_documentation_agent.py
```diff
@@ -31,11 +31,43 @@ def create_search_documentation_json_schema(available_section_file_paths: list[s
...
+                        "description": "True if the query is asking about a feature or capability that WorkflowAI does not currently support.",
+                    },
+                    "feedback": {
+                        "type": "string",
+                        "description": "Explanation of what unsupported feature the user was asking about and why it's not available in WorkflowAI.",
+                    },
+                },
+                "description": "Detection of queries about features that WorkflowAI doesn't support (distinct from missing documentation).",
+                "default": None,
+                "required": ["is_unsupported", "feedback"],
```
pierrevalade, line 48: can `feedback` as a string be `null`? I understand that `is_unsupported` should be required, but what about `feedback`?

api/core/agents/search_documentation_agent.py
```diff
@@ -31,11 +31,43 @@ def create_search_documentation_json_schema(available_section_file_paths: list[s
...
     }
 
 
+class UnsupportedFeatureDetection(BaseModel):
+    """Model for detecting unsupported feature queries."""
+
+    is_unsupported: bool = Field(
+        description="True if the query is asking about a feature or capability that WorkflowAI does not currently support.",
+    )
+    feedback: str = Field(
```
pierrevalade, line 61: could be `none` too? I would have assumed the field `feedback` would be optional (when `is_unsupported = false`)

api/api/routers/mcp/_mcp_service.py
pierrevalade: i would prefer that the MCP client makes its own reasoning based on the documentation pages returned. the goal of this PR was not to have any visible client-side changes.

api/api/routers/mcp/_mcp_service.py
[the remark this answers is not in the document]
pierrevalade: the example you show do not "break" the MCP client in anyway. let's try to keep the changes to a minimum.

api/api/routers/mcp/_mcp_service.py
pierrevalade: > I think that if no relevant documentation is found, you need to explain that you are returning the foundations as "default", but that is not really a documentation section relevant to the query.

has not been an issue so far in my own MCP tests.

api/api/routers/mcp/_mcp_service.py
yannbu: Ok, I'll revert that

api/core/agents/search_documentation_agent.py
pierrevalade: ok sounds good.

api/api/routers/mcp/_mcp_service.py
```diff
@@ -375,6 +376,14 @@ async def search_documentation(
             error="Invalid parameters provided",
         )
 
+    def _get_search_documentation_by_query_message(
+        self,
+        relevant_sections: list[DocumentationSection],
+    ) -> str:
+        if len(relevant_sections) > 0:
+            return f"Successfully found relevant documentation sections: {[section.file_path for section in relevant_sections]}."
+        return "No relevant documentation sections found for your query. Please check the 'foundations' page for general information about WorkflowAI."
```
pierrevalade, line 385: sorry but please can you rollback this change? I did not want any changes to the MCP server.

this PR was about adding one field to the output schema of the search documentation agent, nothing more.

I don't have the time to review the impact your change of the copy returned by the MCP server right now.

thank you!

yannbu: reverted

PR #639, opened 2025-06-30
Rename create_completion to test_agent to clarify purpose

pierrevalade:
## Problem

Based on feedback from Max, the `create_completion` tool name was causing confusion in the MCP client. MCP client (Cursor) thought they could use it directly in production code for their applications, when it's specifically designed for testing and debugging purposes only.

CleanShot 2025-07-01 at 00 17 02@2x

## Solution

Renamed the function to `test_agent` which makes the testing/debugging purpose immediately clear and prevents confusion about production usage.

## Changes Made

- **Renamed function**: `create_completion` → `test_agent`
- **Improved descriptions**: Updated function and parameter descriptions for clarity
- **Enhanced `original_run_id` parameter**: Much more detailed description explaining when and how to use it for debugging
- **Added clear warnings**: Explicit section about what the tool is NOT intended for
- **Restructured documentation**: Cleaner `when_to_use` section with focused bullet points
- **Clarified agent_id**: Simplified parameter description

## Context

This addresses the confusion where MCP clients might think `create_completion` is suitable for production integrations. The new name `test_agent` makes it obvious this is a testing/debugging tool, not a production API endpoint.

The improved documentation also provides better guidance on proper usage patterns and alternatives for production use cases.

### DIFF
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -687,10 +687,10 @@ async def search_documentation(
 
 
 @_mcp.tool()
-async def create_completion(
+async def test_agent(
     # TODO: we should not need the agent id here
     agent_id: str = Field(
-        description="The id of the user's agent. Example: 'agent_id': 'email-filtering-agent' in metadata, or 'email-filtering-agent' in 'model=email-filtering-agent/gpt-4o-latest'.",
+        description="The id of the agent. Example: 'agent_id': 'email-filtering-agent' in metadata, or 'email-filtering-agent' in 'model=email-filtering-agent/gpt-4o-latest'.",
     ),
     # TODO: we should likely split the completion request object
     request: OpenAIProxyChatCompletionRequest = Field(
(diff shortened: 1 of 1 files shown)


api/api/routers/mcp/mcp_server.py
```diff
@@ -687,32 +687,34 @@ async def search_documentation(
 
 
 @_mcp.tool()
-async def create_completion(
+async def test_agent(
```
pierrevalade, line 690: we could also implement `retry_run` like https://github.com/WorkflowAI/WorkflowAI/pull/581 as a alternative option, but I wanted to keep the option for passing a completion without a `original_run_id` so I'm trying this approach first.

guillaq (approved):

api/api/routers/mcp/mcp_server.py
```diff
@@ -687,32 +687,34 @@ async def search_documentation(
...
+    Use test_agent to:
     - Test or compare different AI models without local setup
-    - Create a completion for a WorkflowAI agent (new or existing)
-    - Quickly prototype prompts, structured outputs, or templates
     - Debug agent behavior by testing specific inputs
+    - Quickly prototype prompts, structured outputs, or templates
     - Compare model performance (speed, cost, quality)
     - Retry an existing run with different parameters (requires 'original_run_id' to be provided)
 
+    ⚠️ IMPORTANT: This tool is NOT intended for:
```
guillaq, line 714: Should we add something like `use ... instead` ?

guillaq:
Super weird that cursor would suggest to call an MCP tool server side

PR #651, opened 2025-07-01
Truncate fields above 1000 char long

guillaq:
Quick potential fix for https://linear.app/workflowai/issue/WOR-5117/mcp-search-runs-tool-fails-with-single-item-exceeds-token-limit-error

- limit message base64 data at 10 characters since it is not really interesting for LLMs
- limit input values to 1000 characters
- limit message text content to 10000 characters

Size is arbitrary for now. Just to test what is actually helpful for the LLM

Truncated text gets the `...Truncated` suffix

For now, there is no mention of a way to "fetch the entire payload".

### DIFF
--- a/api/api/routers/mcp/_mcp_errors.py
+++ b/api/api/routers/mcp/_mcp_errors.py
@@ -2,7 +2,6 @@ import json
 from collections.abc import Callable, Coroutine
 from typing import Any
 
-from api.routers.mcp._mcp_models import MCPToolReturn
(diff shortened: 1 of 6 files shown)


guillaq:
> giving a way to the MCP client to get the untruncated payload, maybe via fetch_run_details ?

I do not truncate by default on `fetch_run_details` now

pierrevalade:
ok, noted.

PR #658, opened 2025-07-02
Output schema edition mock

yannbu:


yannbu:
@jacekzimonski once this is merged you will have to trigger the tool with "Can you add description to the output schema fields ?" for example

PR #664, opened 2025-07-03
Integrate mcp playground

guillaq:


api/api/routers/mcp/mcp_server.py
```diff
@@ -803,6 +805,65 @@ async def get_code(
     )
 
 
+@_mcp.tool()
+async def playground(
+    models: str = Field(description="The models to use for the playground (comma-separated)"),
+    lists_of_messages: list[list[OpenAIProxyMessage]] = Field(
+        description="One or several list of messages to use for the playground. Messages can be templated using Jinja2 templates. When using prompt templating, all lists of messages but accept the same set of variables.",
```
pierrevalade, line 812: One or more lists of messages to use in the playground. Messages can be templated using Jinja2 templates. When using prompt templating, all message lists must accept the same set of variables.

PR #666, opened 2025-07-04
07 03 Thoughts

guillaq:
A thought exercise on what WorkflowAI could be. To be taken with a grain of salt.

### DIFF
--- a/.discussions/07_03_guillaume_thoughts.md
+++ b/.discussions/07_03_guillaume_thoughts.md
@@ -16,7 +16,10 @@ So if we apply the principles of Dev UX to how we have built WorkflowAI, we made
 
 So to rethink WorkflowAI as (basically) an SDK, we need to think about a very limited number of core primitives that can be organized by a developer.
 
-Another thing that's interesting is that all the actions that can be performed on primitives should be possible via the UI, code and MCP in some form or another. I don't think it's a matter of **what action can be performed where**, but rather **how can the action be optimized for the different mediums**. For example, the "edit output" action is just **not part of WorkflowAI** period. It is the responsibility of the developer. Our responsibility is to provide the dev with the right tool and documentation.
+Another thing that's interesting is that all the actions that can be performed on primitives should be possible via the UI, code and MCP in some form or another. I don't think it's a matter of **what action can be performed where**, but rather **how can the action be optimized for the different mediums**. For example:
+
+- the "edit output" action is just **not part of WorkflowAI** period. It is the responsibility of the developer. Our responsibility is to provide the dev with the right tool and documentation.
+- the playground should be available via the UI and the MCP (it is already available via code because it's basically a for loop of the completion call). Note that they don't need to have the same shape, running a playground accross 3 different dimensions would be REALLY hard in the UI. Also an "experiment" in the UI should contain the entirety of a user session, not just a run on 3 models. In the UI, the layout should probably resemble the "run" detail view for clarity.
 
 # Primitives:
 


.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+- viewing a completion
+- searching for completions (maybe manipulate the object via a raw SQL query, i.e., group by input, model, etc.)
+- annotate a completion
+
+A completion includes:
+
+- request parameters (aka Versions): templated or system prompt, model, input variables, response format
+- output: response, tool calls
+- raw messages
+- annotations (maybe we can start with a single annotation, and later have annotations per dimension (aka insight)). An annotation has a numerical value (0-1), with 0 being completely negative and 1 being completely positive.
```
pierrevalade, line 40: a `feedback(string)` comment (without any score) is important too.

.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Versions
+
+A version is a set of request parameters used to create a completion.
+
+> I am not 100% sure that a version should be anything more than "part of a completion". For example, do we actually need to save versions?
```
pierrevalade, line 46: I agree.

.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Leave written feedback on runs and have AI use the feedback to improve prompt (Xavier)
+
+> That's the basic concept of annotations. Viewing runs in YAML and being able to inline comments would be insanely good.
+
+1. (Cursor) "Checking all runs with annotations, can you see if you can improve the prompt?"
+2. (Cursor) fetches all the runs with annotations, builds a new prompt and runs an experiment. Cursor can annotate each new run automatically
+3. (UI) runs and annotations are visible in an experiment
+
+> It is worth trying to have Cursor do the improve prompt itself. I am wondering if we will not run into context length / consistency issues.
```
pierrevalade, line 139: we might want to try both approaches.

.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+- we went back with it for a very brief time using the "agent builder" pattern
+- we made the python SDK capable of auto-creating agents for that purpose
+  etc.
+
+But that was never satisfactory because we could not settle on a nice UX. MCP solves that problem and makes it possible.
+
+Which means that now, we don't really need UX anymore. Only developer experience.
+So we need to think about WorkflowAI as it would be used by developers, and by developers only. A developer can possibly be an agent.
+
+So if we apply the principles of Dev UX to how we have built WorkflowAI, we made the mistake of thinking of features that are too high level and don't offer enough flexibility. Reviews are a good example. They are computed auto-magically and hide the concept of datasets.
```
pierrevalade, line 15: agreed.

.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+
+So to rethink WorkflowAI as (basically) an SDK, we need to think about a very limited number of core primitives that can be organized by a developer.
+
+Another thing that's interesting is that all the actions that can be performed on primitives should be possible via the UI, code and MCP in some form or another. I don't think it's a matter of **what action can be performed where**, but rather **how can the action be optimized for the different mediums**. For example:
```
pierrevalade, line 19:
```suggestion
Another thing that's interesting is that all the actions that can be performed on primitives should be possible via the UI, code, API and MCP in some form or another. I don't think it's a matter of **what action can be performed where**, but rather **how can the action be optimized for the different mediums**. For example:
```

.discussions/07_03_guillaume_thoughts.md
```diff
@@ -0,0 +1,148 @@
...
+
+- the "edit output" action is just **not part of WorkflowAI** period. It is the responsibility of the developer. Our responsibility is to provide the dev with the right tool and documentation.
```
pierrevalade, line 21: what is `edit output`? could you please explain? thanks.

.discussions/07_03_guillaume_thoughts.md
pierrevalade, line 40: thinking more, I wonder if we should introduce the concept of "comments" on basically each single field of a `run`, similar to how one can add a comment anywhere on a YAML file, and similar to Github where one can comment on every single field.

LLMs don't care about structured data, everything is just a token for a LLM. we should embrace this new reality completely.

PR #678, opened 2025-07-09
[WOR-5069] Automatically focus text field when new message is added

jacekzimonski:
Closes:
https://linear.app/workflowai/issue/WOR-5069/automatically-focus-text-field-when-new-message-is-added

Demo:
Screen Recording 2025-07-09 at 11 11 44 AM


anyacherniss (approved):
This passes QA for me 👍 Not sure if the bug bot bugs needs to be fixed, but will approve from a QA perspective, since it works for me as expected.

guillaq:
@jacekzimonski can you take a look at bugbot ? (if needed add a comment in code to explain why it's wrong)

PR #680, opened 2025-07-10
Cerebras provider

yannbu:
Not to merge for now (see https://linear.app/workflowai/issue/WOR-2886/provider-cerebras#comment-2e5b2882)

Related to https://linear.app/workflowai/issue/WOR-2886/provider-cerebras)


yannbu:
I think the failing test is only because the CEREBRAS_API_KEY is not taken into account in the CI, but I've added it.

PR #681, opened 2025-07-10
Add grok 4

yannbu:
Related to https://linear.app/workflowai/issue/WOR-5185/add-grok-4-0709


yannbu:
Screenshot 2025-07-10 at 16 33 27
Grok seems to be returning the reasoning token in a different way than the 3 that creates this weird `thinking... thinking... —but I don't think that's a major problem.

PR #682, opened 2025-07-10
Add Grok 4 0907

yannbu:


yannbu:
Grok seems to be returning the reasoning token in a different way than the 3 that creates this weird `thinking... thinking... —but I don't think that's a major problem.
464778980-1ae2d7b2-0654-4504-bc8c-0126f4b90975

api/core/domain/models/model_provider_data_mapping.py
```diff
@@ -1135,6 +1135,13 @@
 }
 
 XAI_PROVIDER_DATA: ProviderDataByModel = {
+    Model.GROK_4_0709: ModelProviderData(
+        text_price=TextPricePerToken(
+            prompt_cost_per_token=5 * ONE_MILLION_TH,
```
anyacherniss, line 1140: @yannbu is this the input token cost? If so, I think it should be 3, not 5!

CleanShot 2025-07-10 at 10 55 59@2x
https://docs.x.ai/docs/models/grok-4-0709

yannbu:
> @yannbu I've read that the reasoning tokens are not returned by Grok so that is probably why you only see these "thinking..." tokens.

@pierrevalade yes, Grok 4 return multiple thinking.... chunk, whereas Grok 3 (on the right on my screenshot) returns correct thinking chunk.

The question is "do we want to write custom logic for Grok 4 to ignore the thinking?" my guess is no, to not spend time on that matter, and Grok might plug the actual tokens in the future. In this case, we won't have to do anything, and we'll have the reasoning, like for Grok 3.

guillaq (approved):

PR #688, opened 2025-07-14
Add field selector and pagination support to get_run endpoint

pierrevalade:
## Summary
- Add new parameters to the `get_run` MCP tool for better handling of large run data
- Enable pagination within specific fields to avoid token limits
- Support wildcard selectors for extracting multiple matching fields

## Changes
- **field_selector**: Optional parameter to extract specific field(s) using dot notation (e.g., `agent_input.email_body` or `messages[*].content[0].text`)
- **offset**: Character position to start reading from (default: 0)
- **limit**: Maximum characters to return (default: 10000)

## Behavior
- When `field_selector` is specified: Returns only the content of matching field(s)
- When `field_selector` is not specified: Returns full run structure with offset/limit applying to all text fields
- Truncated fields now include metadata: `...Truncated[offset=X,shown=Y,total=Z]`

## Examples
```python
# Get full run with truncation
get_run(agent_id="...", run_id="...", truncate=True)

# Get specific field content with pagination
get_run(agent_id="...", run_id="...", field_selector="agent_input.email_body", offset=1000, limit=5000)

# Get all message texts using wildcard
get_run(agent_id="...", run_id="...", field_selector="messages[*].content[0].text")
```

## Notes
- This PR only updates the MCP server endpoint
- The service layer implementation needs to be updated to support these new parameters
- Backward compatible with existing usage

### DIFF
--- a/api/api/routers/mcp/mcp_server.py
+++ b/api/api/routers/mcp/mcp_server.py
@@ -202,9 +202,43 @@ async def get_run(
         bool,
         Field(
             description="Whether to truncate extra long fields in the run details. "
-            "Set to true if the run details response is too long",
+            "Set to true if the run details response is too long. "
+            "When true, long text fields will be cut off and marked with "
+            "'...Truncated[offset=X,shown=Y,total=Z]' where X is the starting position, "
+            "Y is the number of characters shown, and Z is the total length.",
         ),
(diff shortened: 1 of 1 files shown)


guillaq (approved):
Code looks good !

